PR #3, opened 2023-04-27
feat: iam request definition

sqin2019:
gcp policy type: https://cloud.google.com/resource-manager/reference/rest/Shared.Types/Policy
will add a sample yaml file documentation in the future.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
...
+      - 'main'
+  workflow_dispatch:
+
+concurrency:
+  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
+  cancel-in-progress: true
+
+jobs:
+  # Linting job - go.
+  go_lint:
```
sethvargo, line 18: Should we hook up `go_test` while we're here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
+name: 'ci'
```
yolocs, line 1: Per our latest practice, we should add copyright header for yaml files.

.github/workflows/ci.yml
sqin2019: Done.

apis/v1alpha1/iam_request.go
```diff
@@ -0,0 +1,33 @@
...
+	ResourcePolicies []*ResourcePolicy `json:"policies,omitempty"`
+}
+
+// ResourcePolicy specifies the IAM policy to be added for a GCP resource.
+type ResourcePolicy struct {
+	// Resource represents one of GCP organization, folder, and project.
+	Resource string `json:"resource,omitempty"`
+
+	// Policy contains a list of IAM members to role bindings.
+	Policy *resourcemanager.Policy `json:"policy,omitempty"`
```
yolocs, line 32: This will expose other fields in the policy, such as condition. I doubt we want to have that as user-provided input?

sqin2019: Yes, if it is about confusing users we can add readme calling out that those fields will be ignored?

yolocs: I'm not sure why we need the whole `Policy` when we only care about role and members?

sqin2019: Just why not use the existing type, but if you prefer the below alternative, I don't have strong opinion on this since it is simple enough.
```
type ResourcePolicy struct {
  Resource string 
  Bindings []*Binding 
}

type Binding struct {
  Members []string 
  Role string 
}
```

yolocs: > why not use the existing type

Because we only need a subset of it and we don't control the existing type - not even its document. The best document is the API (the contract) itself rather than a readme.

Since it's simple to replicate the binding, I'd prefer we do that.

sqin2019: Updated.

apis/v1alpha1/iam_request.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// ResourcePolicy specifies the IAM principals/members to role bindings to be
+// added for a GCP resource IAM policy.
+type ResourcePolicy struct {
+	// Resource represents one of GCP organization, folder, and project.
+	Resource string `json:"resource,omitempty"`
+
+	// Bindings contains a list of IAM principals/members to role bindings.
+	Bindings []*Binding `json:"binding,omitempty"`
```
yolocs, line 31:
```suggestion
	Bindings []*Binding `json:"bindings,omitempty"`
```

yolocs (approved):

apis/v1alpha1/iam_request.go
```diff
@@ -0,0 +1,44 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+// IamRequest represents a request to update IAM policies.
+type IamRequest struct {
```
sethvargo: For human-written code, the preferred name would be `IAMRequest`. Go is more tolerant with the code is auto-generated (which is why you'll see `Iam` in some client libraries).

apis/v1alpha1/iam_request.go
```diff
@@ -0,0 +1,44 @@
...
+	// List of ResourcePolicy, each specifies the IAM principals/members to role
+	// bindings to be added for a GCP resource IAM policy.
+	ResourcePolicies []*ResourcePolicy `json:"policies,omitempty"`
+}
+
+// ResourcePolicy specifies the IAM principals/members to role bindings to be
+// added for a GCP resource IAM policy.
+type ResourcePolicy struct {
```
sethvargo, line 26: I think we can also have "conditions" here?

sethvargo (changes requested):

apis/v1alpha1/iam_request.go
sqin2019: Good to know, updated.

apis/v1alpha1/iam_request.go
sqin2019: This is for parsing the IAM request which only contains roles and members. Condition is for permission expiry, where we would obtain the duration from PR labels according to the design, and is added when we call GCP API to update the policy.

yolocs: How come? We don't want a requester to be able to set expiry themselves.

sethvargo: Conditions aren't just for expiration; they can be used to define context-aware access and other attributes.

yolocs: I'd say leave the condition out for now. If we introduce it now, we need to look into `Expr` https://pkg.go.dev/google.golang.org/api/cloudresourcemanager/v3#Expr and find whether it's a expiry condition and disallow that.
We can either try to support the condition later or leave it to the `gcloud` command request

yolocs: Per offline discussion, we agreed we can leave the condition out at the moment.

apis/v1alpha1/iam_request.go
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+// IAMRequest represents a request to update IAM policies.
+type IAMRequest struct {
+	// List of ResourcePolicy, each specifies the IAM principals/members to role
+	// bindings to be added for a GCP resource IAM policy.
+	ResourcePolicies []*ResourcePolicy `json:"policies,omitempty"`
```
yolocs, line 21: Forgot to mention, do we need `yaml` tag?

sqin2019: Good catch, will change to yaml tag instead!

sethvargo (approved):

PR #5, opened 2023-04-28
(wip) initial idea on the IAM request handler

sqin2019:
Sharing my thoughts @yolocs 
CLI commands to be implemented:
  - validate IAM request
  - execute/handle IAM request - which will call the handler in this PR


pkg/handler/iam_request_handler.go
```diff
@@ -0,0 +1,62 @@
...
+	service *cloudresourcemanager.Service
+}
+
+// a AOD CLI will be responsible for parse/validate the iam yaml file and unmarshal it
+// to IAMRequest, and then a CLI command to call this handler to execute the request.
+func (h *IAMRRequestHandler) IAMRequestHandler(r *v1alpha1.IAMRequest, ttl *durationpb.Duration) error {
+	for _, p := range r.ResourcePolicies {
+		if err := h.handlePolicy(p, ttl); err != nil {
+			return fmt.Errorf("failed to update IAM policy: %w", err)
+		}
```
sethvargo, line 36: Are we sure we want to stop after a failure? Or should we continue and aggregate all errors instead?

pkg/handler/iam_request_handler.go
```diff
@@ -0,0 +1,62 @@
...
+	switch r := p.Resource; r {
+		case "organization":
+			// Get current policy
+			getIAMPolicyRequest := &cloudresourcemanager.GetIamPolicyRequest{}
+			policy, _ := h.service.Organizations.GetIamPolicy(r, getIAMPolicyRequest).Do()
+			// Update policy with what's in the request
+			updatePolicy(policy, p.Bindings, ttl)
+			// Set policy
+			setIAMPolicyRequest := &cloudresourcemanager.SetIamPolicyRequest{}
+			h.service.Organizations.SetIamPolicy(r, setIAMPolicyRequest)
```
sethvargo, line 51: You're going to want to do this in a loop with retries, since IAM policies use etags to prevent concurrent modification.

pkg/handler/iam_request_handler.go
sqin2019: Hey, haven't thinking about the details yet, but thanks for this reminder!

pkg/handler/iam_request_handler.go
yolocs: +1

pkg/handler/iam_request_handler.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"fmt"
+
+	"github.com/abcxyz/access-on-demand/apis/v1alpha1"
+	"google.golang.org/api/cloudresourcemanager/v3"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+type IAMRRequestHandler struct {
```
yolocs, line 26:
```suggestion
type IAMHandler struct {
```

pkg/handler/iam_request_handler.go
yolocs: +1

PR #7, opened 2023-05-09
feat: iam request handling

### LINKED ISSUE #4: Add logic that execute IAM request

sqin2019:
fix #4


pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,242 @@
...
+				}
+			}
+			// Copy the bindings with de-dupped members.
+			if len(ms) != 0 {
+				cb.Members = ms
+				removed = append(removed, cb)
+			}
+		} else {
+			removed = append(removed, cb)
+		}
```
yolocs, line 200: To help reduce nesting, you could handle this first

```go
if cb.Condition == nil || cb.Condition.Title != conditionTitle {
  result = append(result, cb)
  continue
}

File: pkg/handler/iam_handler.go [hunk 2]
```diff
@@ -0,0 +1,242 @@
...
+							ms = append(ms, m)
+						}
+					}
+				}
+			}
+			// Copy the bindings with de-dupped members.
+			if len(ms) != 0 {
+				cb.Members = ms
+				removed = append(removed, cb)
+			}
```
yolocs, line 197: Maybe use a separate func e.g. `cb.Members = newMembers(cb, bs)`

I'm not a fan of 4 layers of nested for loops. I wonder if we can preprocess `nb` so that it becomes a `map[role]map[member]struct{}` so that you can avoid the nesting the inner two for loops

yolocs (changes requested):

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolicy, ttl time.Duration) (*v1alpha1.IAMResponse, error) {
+	var iamC iamClient
+	switch strings.Split(p.Resource, "/")[0] {
+	case "organizations":
+		iamC = h.organizationsClient
+	case "folders":
+		iamC = h.foldersClient
+	case "projects":
+		iamC = h.projectsClient
```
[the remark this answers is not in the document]
sqin2019: Request validation should catch this?

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	var np *iampb.Policy
```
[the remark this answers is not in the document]
sqin2019: new....

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,242 @@
...
+		updatePolicy(cp, p.Bindings, ttl)
+
+		// Set the new policy.
+		setIamPolicyRequest := &iampb.SetIamPolicyRequest{
+			Resource: p.Resource,
+			Policy:   cp,
+		}
+		np, err = iamC.SetIamPolicy(ctx, setIamPolicyRequest)
+		// Retry when set IAM policy fail.
+		if err != nil {
```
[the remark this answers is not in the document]
sqin2019: I am not sure how much deeper I should go look into this error to get a error code. The error code is probably [409](https://cloud.google.com/resource-manager/docs/core_errors#CONFLICT), is it ok to make it a TODO for now?

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,242 @@
...
+	case "organizations":
+		iamC = h.organizationsClient
+	case "folders":
+		iamC = h.foldersClient
+	case "projects":
+		iamC = h.projectsClient
+	}
+
+	var np *iampb.Policy
+	b := retry.WithMaxRetries(5, retry.NewFibonacci(500*time.Millisecond))
```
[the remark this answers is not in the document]
sqin2019: Like what?

pkg/handler/iam_handler.go
sqin2019: `separate func e.g. cb.Members = newMembers(cb, bs)` -> sg
an issue with preprocess nb to a map is that it may contain bindings that have the same role, I like iterating over the list so that we don't assume anything here.

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,242 @@
...
+	for _, a := range s {
+		if a == e {
+			return true
+		}
+	}
+	return false
+}
+
+// Cleanup handles the graceful shutdown of the resource manager clients.
+func (h *IAMHandler) Cleanup() (retErr error) {
```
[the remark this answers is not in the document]
sqin2019: Could you explain why that is better?

pkg/handler/iam_handler.go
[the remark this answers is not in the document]
yolocs: Please make it a TODO.

pkg/handler/iam_handler.go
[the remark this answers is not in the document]
yolocs: I mean allow providing a retry option when creating the handler

apis/v1alpha1/iam_response.go
```diff
@@ -0,0 +1,26 @@
...
+import "cloud.google.com/go/iam/apiv1/iampb"
+
+// IAMResponse contains the IAM policy returned and its resource information.
+type IAMResponse struct {
+	// IAM policy of the resource.
+	Policy *iampb.Policy
+
+	// Resource represents one of GCP organization, folder, and project.
+	Resource string
+}
```
yolocs, line 26: Forgot to ask - How do we plan to make use of this?

pkg/handler/iam_handler.go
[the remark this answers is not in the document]
yolocs: In general, the lifecycle of an object should be controlled on same layer. E.g. If I create a client, I should be the one to close it. In your approach, the clients could be created by the caller but closed by the handler. So we might as well let the caller do the job.

Plus, `resourcemanager.NewXxxClient` is not an interface. Having them handled outside of this package provides a clearer dependency injection.

apis/v1alpha1/iam_response.go
sqin2019: Direct use of it is in unit test, besides it make sense to return the new policy after updating similar to the go client SetIamPolicy which also returns the new policy.

pkg/handler/iam_handler.go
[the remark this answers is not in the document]
sqin2019: I see...

apis/v1alpha1/iam_request_wrapper.go
```diff
@@ -0,0 +1,28 @@
...
+
+package v1alpha1
+
+import "time"
+
+// IAMRequestWrapper wraps the IAMRequest and adds additional fields such as
+// duration.
+type IAMRequestWrapper struct {
+	// Request contains IAM binding information.
+	Request *IAMRequest
```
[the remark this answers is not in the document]
sqin2019: Just to confirm IIUC, you meant the following? Will it confuse people that duration must also be provided in the PR (instead of by the PR label), or you meant to put IAMRequestWrapper and IAMRequest in the same file.
```go
type IAMRequest struct {
	ResourcePolicies []*ResourcePolicy `yaml:"policies,omitempty"`
        Duration time.Duration
}
```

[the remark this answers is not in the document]
sqin2019: discussed offline

pkg/handler/iam_handler.go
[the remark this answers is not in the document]
sqin2019: discussed offline

apis/v1alpha1/iam_response.go
sqin2019: discussed offline

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,209 @@
...
+
+// ConditionTitle of IAM bindings added by AOD.
+var ConditionTitle = "abcxyz-aod-expiry"
+
+// IAMHandler updates IAM policies of GCP organizations, folders, and projects
+// based on the IAM request received.
+type IAMHandler struct {
+	organizationsClient *resourcemanager.OrganizationsClient
+	foldersClient       *resourcemanager.FoldersClient
+	projectsClient      *resourcemanager.ProjectsClient
```
yolocs, line 42: Here you are still depending on the implementation. Instead, I think they should be of type `IAMClient` (your interface)

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,209 @@
...
+	var iamC iamClient
+	switch strings.Split(p.Resource, "/")[0] {
+	case "organizations":
+		iamC = h.organizationsClient
+	case "folders":
+		iamC = h.foldersClient
+	case "projects":
+		iamC = h.projectsClient
+	default:
+		return nil, fmt.Errorf("unable to match none of the organizations, folders, projects resource")
```
yolocs, line 113:
```suggestion
		return nil, fmt.Errorf("resource isn't one of [organizations, folders, projects]")
```

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	var np *iampb.Policy
+	if err := retry.Do(ctx, h.retry, func(ctx context.Context) error {
+		// Get current IAM policy.
+		cp, err := iamC.GetIamPolicy(
+			ctx,
+			&iampb.GetIamPolicyRequest{Resource: p.Resource})
```
yolocs, line 121: nit: seems unnecessary line breaks?
```suggestion
		cp, err := iamC.GetIamPolicy(ctx, &iampb.GetIamPolicyRequest{Resource: p.Resource})
```

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,209 @@
...
+		return nil, fmt.Errorf("failed to handle IAM request: %w", err)
+	}
+
+	return &v1alpha1.IAMResponse{Resource: p.Resource, Policy: np}, nil
+}
+
+// Remove expired bindings and add or update new bindings with expiration condition.
+func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, ttl time.Duration) {
+	// Convert new bindings to a role to bindings map.
+	bsMap := convert(bs)
```
yolocs, line 151:
```suggestion
	bsMap := toBindingsMap(bs)
```

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,209 @@
...
+	m := make(map[string][]string)
+	for _, b := range bs {
+		m[b.Role] = removeCommonValues(m[b.Role], b.Members)
+		m[b.Role] = append(m[b.Role], b.Members...)
+	}
+	return m
+}
+
+// Returns a result list of strings in l1 that are not found in l2.
+func removeCommonValues(l1, l2 []string) []string {
```
yolocs, line 194: Why not use a map? `map[string]struct{}`

pkg/handler/iam_handler_test.go
```diff
@@ -0,0 +1,565 @@
...
+		})
+	}
+}
+
+type fakeOrgsServer struct {
+	resourcemanagerpb.UnimplementedOrganizationsServer
+
+	policy          *iampb.Policy
+	getIAMPolicyErr error
+	setIAMPolicyErr error
```
yolocs, line 503: To slightly reduce the duplication here, you could put these fields in a separate struct. And that separate struct will implement two func: `getIAMPolicy` and `setIAMPolicy`. By using this struct in all the fakes, you can reduce the duplicated logic.

pkg/handler/iam_handler_test.go
```diff
@@ -0,0 +1,565 @@
...
+								Role: "roles/accessapproval.approver",
+							},
+							{
+								Members: []string{
+									"user:test-org-userC@example.com",
+								},
+								Role: "roles/accessapproval.approver",
+								Condition: &expr.Expr{
+									Title:      ConditionTitle,
+									Expression: fmt.Sprintf("request.time < timestamp('%s')", time.Now().Add(1*time.Hour).Format(time.RFC3339)),
```
yolocs, line 155: This is where flakiness will happen. `time.Now()` here could be different from `time.Now()` called above. Need to use a fixed value created before the test cases.

pkg/handler/iam_handler_test.go
```diff
@@ -0,0 +1,565 @@
...
+			// Setup fake clients.
+			fakeOrgClient, err := resourcemanager.NewOrganizationsClient(ctx, option.WithGRPCConn(connOrg))
+			if err != nil {
+				t.Fatalf("creating client for fake at %q: %v", addrOrg, err)
+			}
+			fakeFolClient, err := resourcemanager.NewFoldersClient(ctx, option.WithGRPCConn(connFol))
+			if err != nil {
+				t.Fatalf("creating client for fake at %q: %v", addrFol, err)
+			}
+			fakeProClient, err := resourcemanager.NewProjectsClient(ctx, option.WithGRPCConn(connPro))
```
yolocs, line 457: These are some weird names. Could you make them normal?

pkg/handler/iam_handler_test.go
```diff
@@ -0,0 +1,565 @@
...
+				resourcemanagerpb.RegisterOrganizationsServer(s, tc.orgsServer)
+			})
+			defer connOrg.Close()
+			addrFol, connFol := testutil.FakeGRPCServer(t, func(s *grpc.Server) {
+				resourcemanagerpb.RegisterFoldersServer(s, tc.foldersServer)
+			})
+			defer connFol.Close()
+			addrPro, connPro := testutil.FakeGRPCServer(t, func(s *grpc.Server) {
+				resourcemanagerpb.RegisterProjectsServer(s, tc.projectsServer)
+			})
```
yolocs, line 445: Do you need one server for each? Can we just register all services to the same gRPC server?

yolocs (changes requested):

pkg/handler/iam_handler.go
sqin2019: Could you provide more info?

yolocs: IIUC, the reason you need to remove common values is just to deduplicate? And you do that with a map

sqin2019: I see, so map[string]struct{} is a set in golang. Sure, updating.

pkg/handler/iam_handler_test.go
sqin2019: When registered the grpc conn, it got confused with which set/getIAMPolicy to use, so still need to define the set/getIAMPolicy for the server struct, have to figure out some other ways if any...

pkg/handler/iam_handler_test.go
sqin2019: Looks like we can! Thx!

pkg/handler/iam_handler_test.go
sqin2019: Tried to put all unimplemented org/folder/projects server in the same struct, since there is a common TestIamPermissions API in all three servers, and it got confused which one to use. To avoid that we have to define TestIamPermissions, but I don't think it is a good idea.

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,202 @@
...
+	organizationsClient IAMClient
+	foldersClient       IAMClient
+	projectsClient      IAMClient
+	// Optional retry backoff strategy, default is 5 attempts with fibonacci
+	// backoff that starts at 500ms.
+	retry retry.Backoff
+}
+
+// Internal IAMClient interface that get and set IAM policies for GCP
+// organizations, folders, and projects.
```
yolocs, line 47:
```suggestion
// IAMClient is the interface to get and set IAM policies for GCP
// organizations, folders, and projects.
```

pkg/handler/iam_handler_test.go
yolocs: I think after you use `IAMClient` in your handler, we don't need separated fakes. We can just need to pick one, say `fakeProjectServer`, it will match the interface for all resources.

sqin2019: Discussed offline.

pkg/handler/__debug_bin
yolocs, line 1: what's this

pkg/handler/iam_handler.go
yolocs: I was wonder whether it'd be easier if `toBindingsMap` returns `map[string]map[string]struct{}`. You don't need to "removeCommonValues" if you're using a map, right?

pkg/handler/iam_handler_test.go
```diff
@@ -0,0 +1,528 @@
...
+package handler
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"cloud.google.com/go/iam/apiv1/iampb"
+	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
```
yolocs, line 25: nit: import order

pkg/handler/iam_handler_test.go
```diff
@@ -0,0 +1,528 @@
...
+}
+
+func setupFakeClients(t *testing.T, ctx context.Context, s1, s2, s3 *fakeServer) (c1, c2, c3 IAMClient) {
+	t.Helper()
+
+	ss := []*fakeServer{s1, s2, s3}
+	cs := make([]IAMClient, 3)
+	for i, e := range ss {
+		// Setup fake servers.
+		addr, conn := testutil.FakeGRPCServer(t, func(s *grpc.Server) {
```
yolocs, line 492: Need to `t.Cleanup` to close these connections

yolocs (commented):
some nit. looking good

pkg/handler/iam_handler.go
sqin2019: removeCommonValues is also used in exclude common Members from current bindings found in the new bindings, see line 158. Is it still worth to change toBindingsMap?

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,201 @@
...
+		// Skip non-AOD bindings.
+		if cb.Condition == nil || cb.Condition.Title != ConditionTitle {
+			result = append(result, cb)
+			continue
+		}
+		// TODO (#6): Remove expired bindings.
+		// Exclude duplicative Members from current bindings.
+		var nm []string
+		for _, m := range cb.Members {
+			if _, ok := bsMap[cb.Role][m]; !ok {
```
yolocs, line 160: There is no guarantee `bsMap[cb.Role]` is in the map in the first place? Should we check it the first thing in the for loop?

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,201 @@
...
+			ms = append(ms, m)
+		}
+		newBinding := &iampb.Binding{
+			Condition: &expr.Expr{
+				Title:      ConditionTitle,
+				Expression: fmt.Sprintf("request.time < timestamp('%s')", t),
+			},
+			Members: ms,
+			Role:    r,
+		}
```
yolocs, line 185:
```suggestion
		newBinding := &iampb.Binding{
			Condition: &expr.Expr{
				Title:      ConditionTitle,
				Expression: fmt.Sprintf("request.time < timestamp('%s')", t),
			},
			Role:    r,
		}
		for m := range members {
			newBinding.Members = append(newBinding.Members, m)
		}
```

yolocs (changes requested):

pkg/handler/iam_handler.go
sqin2019: Done.

pkg/handler/iam_handler.go
```diff
@@ -0,0 +1,203 @@
...
+		if cb.Condition == nil || cb.Condition.Title != ConditionTitle {
+			result = append(result, cb)
+			continue
+		}
+		// TODO (#6): Remove expired bindings.
+		// Keep members that are not in the new bindings.
+		if bsMap[cb.Role] == nil {
+			result = append(result, cb)
+			continue
+		}
```
yolocs, line 161: Please do

```suggestion
		// TODO (#6): Remove expired bindings.
		// Skip roles we are not interested in.
		if _, ok := bsMap[cb.Role]; !ok {
			result = append(result, cb)
			continue
		}
```

pkg/handler/iam_handler_test.go
```diff
@@ -0,0 +1,563 @@
...
+		projectsServer          *fakeServer
+		request                 *v1alpha1.IAMRequestWrapper
+		wantPolicies            []*v1alpha1.IAMResponse
+		wantErrSubstr           string
+		wantOrganizationsPolicy *iampb.Policy
+		wantFoldersPolicy       *iampb.Policy
+		wantProjectsPolicy      *iampb.Policy
+	}{
+		{
+			name: "success",
```
yolocs, line 55: Do you mind splitting this case up? 1. happy path 2. ignore unrelated roles 3. ignore non-AOD bindings 4. clean up duplicated members. This would us to easily identify which case we failed if we changed any logic.

yolocs (approved):

PR #12, opened 2023-05-18
feat: cli handle iam request

### LINKED ISSUE #11: CLI for handling IAMRequest

sqin2019:
fix #11


pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,155 @@
...
+var _ cli.Command = (*IAMHandleCommand)(nil)
+
+type IAMHandleCommand struct {
+	cli.BaseCommand
+
+	flagPath string
+
+	flagDuration time.Duration
+
+	handler *handler.IAMHandler
```
sqin2019, line 40: This is for testing purpose to pass in the handler.

pkg/cli/iam_handle_test.go
```diff
@@ -0,0 +1,261 @@
...
+	pkgtestutil "github.com/abcxyz/pkg/testutil"
+)
+
+func TestIAMHandleCommand(t *testing.T) {
+	t.Parallel()
+
+	// Set up IAM request file.
+	requestFileContentByName := map[string]string{
+		"valid.yaml": `
+policies:
```
sqin2019, line 43: I believe the order in yaml are respected when unmarshalled to IAMRequest (struct of list of struct that contains list), also confirmed it with Bard, but somehow the CI failed because of order ( it is reran so the failed history was gone).

pkg/iam/interfaces.go
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// iam package provides interfaces around the google resourcemanager
+// organizations, folders, projects client to handles IAM policy update.
+package iam
```
sqin2019, line 17: Moved the interface into this new package, if it stays in the handler package, I cannot depend on the testutil since testutil also uses this interface (import cycle). Is this the proper way to do?

pkg/cli/iam_handle.go
sethvargo: Prefix the name with `test` so it's clear that it's for testing.

pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// Wrap IAMRequest to include Duration.
+	reqWrapper := &v1alpha1.IAMRequestWrapper{
+		IAMRequest: req,
+		Duration:   c.flagDuration,
+	}
+	if _, err := c.handler.Do(ctx, reqWrapper); err != nil {
+		return fmt.Errorf("failed to handle IAM request: %w", err)
+	}
+	c.Outf("IAM request handled successfully")
```
sethvargo, line 152:
```suggestion
	c.Outf("Successfully handled IAM request")
```

pkg/iam/interfaces.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+
+	"cloud.google.com/go/iam/apiv1/iampb"
+	"github.com/googleapis/gax-go/v2"
+)
+
+// IAMClient is the interface to get and set IAM policies for GCP organizations,
+// folders, and projects.
+type IAMClient interface {
```
sethvargo, line 28: Since this is now in a package named "iam", we should just rename this "Client".

pkg/testutil/iam_client.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, s.SetIAMPolicyErr
+	}
+	s.Policy = r.Policy
+	return s.Policy, s.SetIAMPolicyErr
+}
+
+func SetupFakeClients(t *testing.T, ctx context.Context, s1, s2, s3 *FakeServer) (c1, c2, c3 iam.IAMClient) {
+	t.Helper()
+
+	ss := []*FakeServer{s1, s2, s3}
```
sethvargo, line 60: This feels arbitrary? Why are we creating 3 servers?

sqin2019: We need three clients in the handler [struct](https://github.com/abcxyz/access-on-demand/blob/cfa52cbfcbbb61adf3194ff2057c1f1c9f3a6278/pkg/handler/iam_handler.go#L38). These are organizations/folders/projects clients in the [go client lib for resourcemanager](https://github.com/googleapis/google-cloud-go/tree/main/resourcemanager/apiv3). An alternative is to just create one server, but we call it three times.

pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/abcxyz/access-on-demand/pkg/handler"
+	"github.com/abcxyz/pkg/cli"
+	"gopkg.in/yaml.v3"
+
+	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
+)
+
+var _ cli.Command = (*IAMHandleCommand)(nil)
+
+type IAMHandleCommand struct {
```
yolocs, line 33: nit: comment throughout

pkg/cli/iam_handle.go
yolocs, line 40: I'd prefer this to be an interface so this package will be decoupled from `IAMHandler`

pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,155 @@
...
+	return `Handle the IAM request YAML file in the given path`
+}
+
+func (c *IAMHandleCommand) Help() string {
+	return `
+Usage: {{ COMMAND }} [options]
+
+Handle the IAM request YAML file in the given path:
+
+      aod iam handle -path "/path/to/file" -duration "2h"
```
yolocs, line 53: I realized that just `duration` might be problematic. Say I changed the duration from 2h to 4h. But it shouldn't change my "start time", right? Maybe we need an optional `-start-time` that's default to `now`.

pkg/cli/iam_handle_test.go
```diff
@@ -0,0 +1,261 @@
...
+	cases := []struct {
+		name                    string
+		args                    []string
+		fileData                []byte
+		organizationsServer     *testutil.FakeServer
+		foldersServer           *testutil.FakeServer
+		projectsServer          *testutil.FakeServer
+		wantOrganizationsPolicy *iampb.Policy
+		wantFoldersPolicy       *iampb.Policy
+		wantProjectsPolicy      *iampb.Policy
```
yolocs, line 90: These are already tested in `IAMHandler` so you shouldn't test them again here.

pkg/iam/interfaces.go
yolocs: I believe if you decouple `IAMHandler` from the command, you wouldn't need this change.

yolocs (changes requested):

pkg/cli/iam_handle.go
sqin2019: Assume the duration change is a label change on the PR which will trigger a different workflow run, we will need to pass down the start time to the next workflow? Or the start time is just for running it locally?

pkg/cli/iam_handle.go
sqin2019: hum.....ok, this will make the test easier then just implement a fake Do. let me try it...

pkg/cli/iam_handle.go
yolocs: I think the start time should be the PR (aka. the request) creation time

sqin2019: Taking a step back, when the iam handle workflow will be triggered? when an IAM request is approved? So I guess the start time need to be the approved time instead. When duration label is changed, will it undo the previous approval?

yolocs: Good point! Label won't undo approval

pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	return c.handleIAM(ctx)
+}
+
+func (c *IAMHandleCommand) handleIAM(ctx context.Context) error {
+	// Unmarshal the YAML file to IAMRequest
+	data, err := os.ReadFile(c.flagPath)
+	if err != nil {
+		return fmt.Errorf("failed to read file from %q, %w", c.flagPath, err)
```
yolocs, line 112: nit
```suggestion
		return fmt.Errorf("failed to read IAM request file at %q: %w", c.flagPath, err)
```

pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,166 @@
...
+			return fmt.Errorf("failed to create IAM handler: %w", err)
+		}
+	}
+
+	// Wrap IAMRequest to include Duration.
+	reqWrapper := &v1alpha1.IAMRequestWrapper{
+		IAMRequest: req,
+		Duration:   c.flagDuration,
+	}
+	if _, err := h.Do(ctx, reqWrapper); err != nil {
```
yolocs, line 160: Could be a follow-up. Add a `--log-level` flag. If set to debug, output the response.

pkg/cli/iam_handle_test.go
```diff
@@ -0,0 +1,141 @@
...
+	for name, content := range requestFileContentByName {
+		path := filepath.Join(dir, name)
+		if err := os.WriteFile(path, []byte(content), 0o600); err != nil {
+			t.Fatal(err)
+		}
+		t.Cleanup(func() {
+			if err := os.Remove(path); err != nil {
+				t.Fatal(err)
+			}
+		})
```
yolocs, line 70: Per https://pkg.go.dev/testing#T.TempDir, the dir will be auto removed after the test.

pkg/cli/iam_handle_test.go
```diff
@@ -0,0 +1,141 @@
...
+				t.Errorf("Process(%+v) got output diff (-want, +got):\n%s", tc.name, diff)
+			}
+		})
+	}
+}
+
+type fakeHandler struct{}
+
+func (h *fakeHandler) Do(context.Context, *v1alpha1.IAMRequestWrapper) ([]*v1alpha1.IAMResponse, error) {
+	return nil, nil
```
yolocs, line 140: Test when handler returns err?

pkg/cli/iam_handle.go
sqin2019: Create an issue to add start time: https://github.com/abcxyz/access-on-demand/issues/13

pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,166 @@
...
+	if c.flagDuration <= 0 {
+		return fmt.Errorf("a positive duration is required")
+	}
+
+	return c.handleIAM(ctx)
+}
+
+func (c *IAMHandleCommand) handleIAM(ctx context.Context) error {
+	// Unmarshal the YAML file to IAMRequest
+	data, err := os.ReadFile(c.flagPath)
```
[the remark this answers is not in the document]
sqin2019: Make senses, thanks! created an issue: https://github.com/abcxyz/access-on-demand/issues/14

pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,166 @@
...
+		h, err = handler.NewIAMHandler(
+			ctx,
+			organizationsClient,
+			foldersClient,
+			projectsClient,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to create IAM handler: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
yolocs: I can't recall precisely because there could be a catch. @sqin2019 I remember you did some research earlier, right?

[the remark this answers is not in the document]
sqin2019: The primary reason is to limit the capability, we only want org/folder/project level iam changes. On google [guide](https://cloud.google.com/iam/docs/granting-changing-revoking-access) also recommends the resource manager clients. --> "You can manage access to projects, folders, and organizations with the Google Cloud console, the Google Cloud CLI, the REST API, or the [Resource Manager client libraries](https://cloud.google.com/resource-manager/docs/libraries)."

It is also doable with the generic IAM client since we already limited the resource type in code. Let me know if we should change to this one instead.

pkg/cli/iam_handle.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/abcxyz/pkg/cli"
+	"gopkg.in/yaml.v3"
+
+	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
+)
+
+var _ cli.Command = (*IAMHandleCommand)(nil)
+
+// Handler interface that handles the IAMRequestWrapper.
+type Handler interface {
```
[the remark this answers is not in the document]
sqin2019: Updated

pkg/testutil/iam_client.go
sqin2019: close dup, will discuss it here: https://github.com/abcxyz/access-on-demand/pull/12#discussion_r1198989681

pkg/cli/iam_handle.go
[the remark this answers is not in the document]
yolocs: Let's talk offline.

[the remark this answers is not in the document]
sqin2019: Discussed with Chen, AI: I will try the generic IAM client and see it works for our use cases, and then follow up on this.

pkg/cli/iam_handle_test.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type fakeFailureHandler struct{}
+
+func (h *fakeFailureHandler) Do(context.Context, *v1alpha1.IAMRequestWrapper) ([]*v1alpha1.IAMResponse, error) {
+	return nil, fmt.Errorf("always fail")
+}
```
yolocs, line 156: Instead of another struct, `fakeHandler` could be reused

```go
type fakeIAMHandler struct {
  injectErr error // return this error in Do
  gotReq *v1alpha1.IAMRequestWrapper // record it from Do and compare it to a 'want'
}

yolocs (changes requested):

yolocs (approved):

pkg/cli/iam_handle.go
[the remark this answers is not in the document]
sqin2019: https://github.com/abcxyz/access-on-demand/issues/17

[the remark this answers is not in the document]
sqin2019: @sethvargo Should I resolve or if any further concerns?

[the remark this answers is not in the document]
sqin2019: I will close and follow up in the issue instead, thanks!

PR #18, opened 2023-05-22
feat: add validate cli

sqin2019:


pkg/requestutil/request.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/abcxyz/access-on-demand/apis/v1alpha1"
+	"gopkg.in/yaml.v3"
+)
+
+func ReadFromPath(path string) (*v1alpha1.IAMRequest, error) {
```
sethvargo, line 27: Doc

sethvargo (approved):

pkg/requestutil/request.go
sqin2019: Oops, updating!

sethvargo:
@sqin2019 if you'd like, you can:

```go
go get -u github.com/abcxyz/pkg@main
```

and then add tab completions. This will be really helpful for browsing to files (you can use `predict.Files` to have tab completion for the local filesystem). See https://github.com/abcxyz/pkg/pull/125 for some examples.

sqin2019:
> @sqin2019 if you'd like, you can:
> 
> ```go
> go get -u github.com/abcxyz/pkg@main
> ```
> 
> and then add tab completions. This will be really helpful for browsing to files (you can use `predict.Files` to have tab completion for the local filesystem). See [abcxyz/pkg#125](https://github.com/abcxyz/pkg/pull/125) for some examples.
> 
> This can also be a follow-up if you want. We just added this functionality, so it's not a blocker.

Thanks, will check it out and follow up!

PR #23, opened 2023-05-23
feat: add start time

### LINKED ISSUE #13: Add start time in IAMRequestWrapper
The start time is PR approved time.

sqin2019:
fix #13

### DIFF
--- a/pkg/cli/iam_handle.go
+++ b/pkg/cli/iam_handle.go
@@ -42,7 +42,7 @@ type IAMHandleCommand struct {
 
 	flagDuration time.Duration
 
-	// TODO(#13): add start time flag.
+	flagStartTime time.Time
 
 	// testHandler is used for testing only.
 	testHandler iamHandler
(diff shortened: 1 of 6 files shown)


pkg/cli/iam_handle.go
```diff
@@ -82,9 +82,33 @@ func (c *IAMHandleCommand) Flags() *cli.FlagSet {
...
+	})
+
 	return set
 }
 
+// Parse the string representation of time, it must be in RFC3339 format.
+func timeParser(s string) (time.Time, error) {
+	t, err := time.Parse(time.RFC3339, s)
+	if err != nil {
+		return t, fmt.Errorf("failed to parse the time with RFC3339 layout: %w", err)
```
yolocs, line 102:
```suggestion
		return t, fmt.Errorf("failed to parse %q with RFC3339 layout: %w", s, err)
```

yolocs (approved):

pkg/cli/iam_handle.go
```diff
@@ -82,9 +82,33 @@ func (c *IAMHandleCommand) Flags() *cli.FlagSet {
 		Usage:   `The IAM permission lifecycle, as a duration.`,
 	})
 
+	cli.Flag(f, &cli.Var[time.Time]{
```
sethvargo, line 85: Good use of the flags package! Do you think we should move this into pkg/cli?

sqin2019: Was thinking about the same thing! created an [issue](https://github.com/abcxyz/pkg/issues/127).

PR #26, opened 2023-05-25
feat: do workflow - IAM track

### LINKED ISSUE #19: GH Do workflow - IAM track

sqin2019:
fix #19 
tested by trigger on push on branch

### DIFF
--- a/.github/workflows/iam-handle.yml
+++ b/.github/workflows/iam-handle.yml
@@ -1,4 +1,8 @@
 name: 'aod-iam-handle'
+# TODO(#30): Support pull_request labeled triggering event. Current workflow
+# requires the triggering event must be pull_request_review as it relies on this
+# event to get the review status and approval submit time as IAM permission
+# expiration start time.
 on:
   workflow_call:
     inputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,58 @@
...
+    inputs:
+      workload_identity_provider:
+        description: 'The full identifier of the Workload Identity Provider, including the project number, pool name, and provider name.'
+        type: 'string'
+        required: true
+      service_account:
+        description: 'Email address or unique identifier of the Google Cloud service account for which to generate credentials.'
+        type: 'string'
+        required: true
+      path:
```
yolocs, line 13: This should be a fixed value? Do we need it to be an input?

.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,58 @@
...
+        type: 'string'
+        required: true
+      start_time:
+        description: 'The start time of the IAM permission lifecycle.'
+        type: 'string'
+        required: true
+      duration:
+        description: 'The IAM permission lifecycle, as a duration.'
+        type: 'string'
+        required: true
```
yolocs, line 24: Shouldn't we detect them automatically? We don't want workflow users to do that.

.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: 'Authenticate to Google Cloud'
+        id: 'auth'
+        uses: 'google-github-actions/auth@35b0e87d162680511bf346c299f71c9c5c379033' # ratchet:google-github-actions/auth@v1
+        with:
+          workload_identity_provider: '${{ inputs.workload_identity_provider }}'
+          service_account: '${{ inputs.service_account }}'
+          token_format: 'access_token'
+      - id: 'install-aodctl'
+        shell: 'bash'
+        run: 'go install github.com/abcxyz/access-on-demand/cmd/aod@latest'
```
yolocs, line 55: This version (current value `latest`) should be a configurable input.

.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          go-version: ${{ inputs.go_version }}
+      - name: 'Authenticate to Google Cloud'
+        id: 'auth'
+        uses: 'google-github-actions/auth@35b0e87d162680511bf346c299f71c9c5c379033' # ratchet:google-github-actions/auth@v1
+        with:
+          workload_identity_provider: '${{ inputs.workload_identity_provider }}'
+          service_account: '${{ inputs.service_account }}'
+          token_format: 'access_token'
+      - id: 'install-aodctl'
```
yolocs, line 53: You're not using input/output from previous steps. I usually prefer name instead like "Install AOD CLI"

.github/workflows/iam-handle.yml
sqin2019: Ok, we can require the path to be ${{ github.workspace }}/iam.yaml

.github/workflows/iam-handle.yml
sqin2019: Was thinking the other way around, make sense, will update.

.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,70 @@
+name: 'aod-iam-handle'
```
yolocs, line 1: License header

.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@35b0e87d162680511bf346c299f71c9c5c379033' # ratchet:google-github-actions/auth@v1
+        with:
+          workload_identity_provider: '${{ inputs.workload_identity_provider }}'
+          service_account: '${{ inputs.service_account }}'
+          token_format: 'access_token'
+      - name: 'Install AOD CLI'
+        run: 'go install github.com/abcxyz/access-on-demand/cmd/aod@${{ inputs.aod_cli_version }}'
+      - name: 'Get Duration From Label'
+        shell: 'bash'
```
yolocs, line 55: Remove?

.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,70 @@
...
+              IFS='-' read -r part1 part2 <<< "$name"
+              echo "DURATION=$part2" >> $GITHUB_ENV
+              break
+            fi
+          done
+      - name: 'Run AOD CLI'
+        shell: 'bash'
+        env:
+          FILE_PATH: ${{ github.workspace }}/iam.yaml
+          START_TIME: ${{ github.event.review.submitted_at }}
```
yolocs, line 69: I think this is a good side effect. When there is new change, branch protection will require re-review. The re-review approval time would be the start time.

.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          workload_identity_provider: '${{ inputs.workload_identity_provider }}'
+          service_account: '${{ inputs.service_account }}'
+          token_format: 'access_token'
+      - name: 'Install AOD CLI'
+        run: 'go install github.com/abcxyz/access-on-demand/cmd/aod@${{ inputs.aod_cli_version }}'
+      - name: 'Get Duration From Label'
+        shell: 'bash'
+        run: |
+          names='${{ toJson(github.event.pull_request.labels.*.name) }}'
```
yolocs, line 57: What if there is no duration label? We should have a default?

yolocs (changes requested):

.github/workflows/iam-handle.yml
sqin2019: Done.

.github/workflows/iam-handle.yml
sqin2019: It should be required here since it runs bash, I removed the one in 'Install AOD CLI'.

.github/workflows/iam-handle.yml
sqin2019: Sure, updating.

.github/workflows/iam-handle.yml
sqin2019: It falls back to sh if no bash is found, tried locally and it is working.

.github/workflows/iam-handle.yml
sqin2019: Made it an input with default.

.github/workflows/iam-handle.yml
sqin2019: Another option is to make it an input with default, not a big deal at this moment, it is going to be a trivial update if required later.

.github/workflows/iam-handle.yml
sqin2019: Changed my mind since we may want to limit the flexibility...

.github/workflows/iam-handle.yml
```diff
@@ -0,0 +1,84 @@
...
+          for name in $(echo "$names" | jq -r '.[]'); do
+            if [[ $name == duration-* ]]; then
+              IFS='-' read -r part1 part2 <<< "$name"
+              echo "LABELED_DURATION=$part2" >> $GITHUB_ENV
+              break
+            fi
+          done
+      - name: 'Run AOD CLI'
+        env:
+          DURATION: ${{ env.LABELED_DURATION ||'2h' }}
```
yolocs, line 81:
```suggestion
          DURATION: ${{ env.LABELED_DURATION || '2h' }}
```

yolocs (approved):

PR #32, opened 2023-05-31
feat: add validate workflow

sqin2019:
Tested with push to issue branch event.


.github/workflows/iam-validate.yml
```diff
@@ -0,0 +1,50 @@
...
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    name: 'Validate IAM Request'
+    steps:
+      - name: 'Checkout Triggering Branch'
+        uses: 'actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab' # ratchet:actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
sethvargo, line 40: Please quote all YAML strings

.github/workflows/iam-validate.yml
```diff
@@ -0,0 +1,50 @@
...
+        required: false
+      go_version:
+        description: 'The version of Golang.'
+        type: 'string'
+        default: '1.20'
+        required: false
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
```
sethvargo, line 34: Please quote all YAML strings

.github/workflows/iam-validate.yml
```diff
@@ -0,0 +1,50 @@
...
+    name: 'Validate IAM Request'
+    steps:
+      - name: 'Checkout Triggering Branch'
+        uses: 'actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab' # ratchet:actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+      - name: 'Setup Go'
+        uses: 'actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568' # ratchet:actions/setup-go@v3
+        with:
+          go-version: ${{ inputs.go_version }}
```
sethvargo, line 44: Please quote all YAML strings

.github/workflows/iam-validate.yml
```diff
@@ -0,0 +1,50 @@
...
+      aod_cli_version:
+        description: 'The version of AOD CLI.'
+        type: 'string'
+        default: 'latest'
+        required: false
+      go_version:
+        description: 'The version of Golang.'
+        type: 'string'
+        default: '1.20'
+        required: false
```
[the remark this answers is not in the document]
sqin2019: Just tested in my own repo, when required is true even with default, and the caller workflow did not provide it, it got an error "Input aod_cli_version is required, but not provided while calling." I think for our use case, we want it to default to the defaults when inputs are not provided.

.github/workflows/iam-validate.yml
sqin2019: Ack!

yolocs: I also keep forgetting. https://github.com/abcxyz/readability/wiki/YAML-Best-Practices

.github/workflows/iam-handle.yml
```diff
@@ -79,6 +79,6 @@ jobs:
       - name: 'Run AOD CLI'
         env:
           DURATION: ${{ env.LABELED_DURATION || '2h' }}
```
sqin2019, line 81: It does not allow me to quote this one because the '2h' already has quote.An alternative is to set a default duration env variable, and do DURATION: ${{ env.LABELED_DURATION || env.DEFAULT_DURATION }}, will check if there is a better option before doing that.

.github/workflows/iam-validate.yml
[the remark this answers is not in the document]
sethvargo: Oh that's super weird...

.github/workflows/iam-validate.yml
sethvargo: We should just write a linter... I wonder if I can nerd snipe @drevell into doing it since he just played around with the YAML ast parser for abc?

.github/workflows/iam-handle.yml
sqin2019: No, it gives me --> Unexpected symbol: '"2h"'. Located at position 25 within expression: env.LABELED_DURATION || "2h"

.github/workflows/iam-validate.yml
sethvargo: No, we'd put this in abcxyz/pkg as part of our linters (kinda like we did with Terraform).

drevell: I could conceivably be sniped into doing that after I ship v1 of the abc CLI 😄

yolocs: LOL reminder created https://github.com/abcxyz/pkg/issues/130

.github/workflows/iam-handle.yml
sqin2019: I am keeping it as is, since DURATION: '${{ env.LABELED_DURATION || env.DEFAULT_DURATION }} ' needs me to define DEFAULT_DURATION in earlier steps. Feel like the current way is the cleanest.

yolocs: Does this work? Define `env.DEFAULT_DURATION` at the top might actually be clearer IMO. And you can avoid introducing an "outlier" here (not quoting).

yolocs (approved):

.github/workflows/iam-handle.yml
sqin2019: Top level should work.

sqin2019: Done.

PR #34, opened 2023-06-07
fix: make the workflows more general and add yamllint

sqin2019:


.github/workflows/handle.yml
```diff
@@ -68,8 +68,15 @@ jobs:
           token_format: 'access_token'
       - name: 'Install AOD CLI'
         run: 'go install github.com/abcxyz/access-on-demand/cmd/aod@${{ inputs.aod_cli_version }}'
-      # Duration labels need to be predixed with "duration-", an example is "duration-2h",
+
+      # Start handling IAM request.
+      - id: 'changed-file-iam'
+        uses: 'tj-actions/changed-files@84ed30e2f4daf616144de7e0c1db59d5b33025e3' # ratchet:tj-actions/changed-files@v35
+        with:
+          files: 'iam.yaml'
```
sqin2019, line 76: Another way is to use hashFiles('iam.yaml') != '' to check the file existence. changed-files is more efficient since it won't handle the file request if no changes to the file, but it requires the file must exists in the triggering branch. Given the file will be created when setting up the repos, it makes sense to use changed-files here.

.github/workflows/ci.yml
```diff
@@ -34,9 +34,11 @@ jobs:
     with:
       go_version: '1.20'
 
+  yaml_lint:
+    uses: 'abcxyz/pkg/.github/workflows/yaml-lint.yml@main' # ratchet:exclude
```
sethvargo, line 38: 🥳

.github/workflows/handle.yml
sethvargo: Why can't we use a constraint on the job to only run when the file is changed? 

```yaml
on:
  push:
    paths:
      - 'iam.yaml'
  pull_request:
    paths:
      - 'iam.yaml'
```

sqin2019: It is because we cannot use path constraint with pull_request_review event. Handle is triggered on this event when review status is approved, [here](https://github.com/abcxyz/aod-template/pull/1/files#diff-898d9cdbbc9e945afcaf482b25e6e81297319146b3887a322faf9368dbb56eabR5) is what the caller workflow will look like.

sethvargo: Ah okay - can we document that really clearly? That wasn't immediately obvious from the YAML or the PR description.

sqin2019: Sure, will add more comment!

yolocs: Does this even work? If `iam.yaml` doesn't exist, does this step fail? I think we should instead update the "Handle IAM Request" step to run:

```sh
if [ -f "iam.yaml" ]; then
  aod iam handle -path $FILE_PATH -duration $DURATION -start-time $START_TIME
else
  echo "::notice title=IAM Request::Skip because iam.yaml is not found"
fi
```

.github/workflows/handle.yml
```diff
@@ -79,9 +86,14 @@ jobs:
...
           done
-      - name: 'Run AOD CLI'
+      - name: 'Handle IAM Request'
+        if: '${{ steps.changed-file-iam.outputs.any_changed == ''true''}}'
         env:
           DURATION: '${{ env.LABELED_DURATION || env.DEFAULT_DURATION }}'
           FILE_PATH: '${{ github.workspace }}/iam.yaml'
           START_TIME: '${{ github.event.review.submitted_at }}'
         run: 'aod iam handle -path $FILE_PATH -duration $DURATION -start-time $START_TIME'
+      - id: 'changed-file-gcloud'
```
yolocs, line 96: Same

.github/workflows/validate.yml
```diff
@@ -44,7 +44,12 @@ jobs:
           go-version: '${{ inputs.go_version }}'
       - name: 'Install AOD CLI'
         run: 'go install github.com/abcxyz/access-on-demand/cmd/aod@${{ inputs.aod_cli_version }}'
-      - name: 'Run AOD CLI'
+      - id: 'changed-file-iam'
+        uses: 'tj-actions/changed-files@84ed30e2f4daf616144de7e0c1db59d5b33025e3' # ratchet:tj-actions/changed-files@v35
+        with:
+          files: 'iam.yaml'
```
yolocs, line 50: Same. I don't understand why we need this.

yolocs (changes requested):

.github/workflows/handle.yml
sqin2019: Discussed offline, the iam.yaml file will only exist when a request is made, therefore we should not use this github action. switching...

.github/workflows/handle.yml
```diff
@@ -68,8 +68,10 @@ jobs:
           token_format: 'access_token'
       - name: 'Install AOD CLI'
         run: 'go install github.com/abcxyz/access-on-demand/cmd/aod@${{ inputs.aod_cli_version }}'
-      # Duration labels need to be predixed with "duration-", an example is "duration-2h",
+      # Duration labels need to be prefixed with "duration-", an example is "duration-2h".
+      # Skip when iam.yaml file does not exist in the case of a pull_request_review event.
       - name: 'Get Duration From Label'
+        if: '${{ hashFiles(''iam.yaml'') != '''' }}'
```
yolocs, line 74: Nice

yolocs (approved):

PR #37, opened 2023-06-07
fix: add condition check in handle workflow

sqin2019:
The condition check helps to skip time consuming steps when a PR is not AOD request.

### DIFF
--- a/.github/workflows/handle.yml
+++ b/.github/workflows/handle.yml
@@ -56,20 +56,24 @@ jobs:
         uses: 'actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab' # ratchet:actions/checkout@v3
         with:
           ref: '${{ github.event.pull_request.head.ref }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/handle.yml
```diff
@@ -56,20 +56,24 @@ jobs:
         uses: 'actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab' # ratchet:actions/checkout@v3
         with:
           ref: '${{ github.event.pull_request.head.ref }}'
+      # Steps will be skipped starting from here when iam.yaml file does not
+      # exist in the case of a pull_request_review event.
       - name: 'Setup Go'
+        if: '${{ hashFiles(''iam.yaml'') != '''' }}'
```
sqin2019, line 62: will add gcloud.yaml once we have it implemented.

yolocs (approved):

yolocs:
I helped to update PR comment. Try to do that for future PRs

PR #39, opened 2023-06-08
fix: add yaml lint in reuable workflow

### LINKED ISSUE #38: Check yaml lint when validate request.

sqin2019:
closes #38

### DIFF
--- a/.github/workflows/handle.yml
+++ b/.github/workflows/handle.yml
@@ -44,6 +44,8 @@ env:
 
 jobs:
   handle:
+    # Only handle the request when the PR is approved.
+    if: '${{ github.event.review.state == ''approved'' }}'
     runs-on: 'ubuntu-latest'
     permissions:
       contents: 'read'


.github/workflows/handle.yml
```diff
@@ -43,9 +43,14 @@ env:
...
 jobs:
-  handle:
+  yaml_lint:
     # Only handle the request when the PR is approved.
     if: '${{ github.event.review.state == ''approved'' }}'
+    uses: 'abcxyz/pkg/.github/workflows/yaml-lint.yml@main' # ratchet:exclude
+
+  handle:
+    # Only run when yaml_lint is successfully completed.
+    needs: 'yaml_lint'
```
sethvargo, line 53: This creates a dependency - do we want to require lint to pass _before_ validation, or should they happen at the same time?

sqin2019: I feel like the dependency is needed here since we want requestor to fix the lint issue before handling the request. For the validate.yml though, we probably can run them at the same time.

.github/workflows/handle.yml
```diff
@@ -43,9 +43,14 @@ env:
   DEFAULT_DURATION: '2h'
 
 jobs:
-  handle:
+  yaml_lint:
     # Only handle the request when the PR is approved.
     if: '${{ github.event.review.state == ''approved'' }}'
+    uses: 'abcxyz/pkg/.github/workflows/yaml-lint.yml@main' # ratchet:exclude
```
sqin2019, line 49: Had to rethink whether we should put this job here since users like aod-template cannot ping the SHA of this workflow. 'uses:' does not have access to github context like envs and inputs. Or using main is fine. @yolocs

yolocs: Let's just pin it here.

.github/workflows/handle.yml
yolocs: Since we already yaml lint in the validate workflow, we don't need to rerun it to cause duplicated checks.

sqin2019: In summary, the yaml lint is just a reference for request reviewer/approver. updating!

bradegler (approved):

PR #41, opened 2023-06-14
feat: add PR expire reusable workflow

sqin2019:


.github/workflows/expire.yml
```diff
@@ -0,0 +1,75 @@
...
+      owner:
+        description: 'The GitHub org or repo owner.'
+        type: 'string'
+        required: true
+      repo:
+        description: 'The repo name.'
+        type: 'string'
+        required: true
+      expiryDays:
+        description: 'The number of days to keep an AOD PR open since last updated.'
```
sethvargo, line 15: "days" feels too large. Should this be hours?

sqin2019: I want to give some buffer for reviewers to approve (an AOD/IAM PR needs approval for AOD to handle the request in the PR. Would hours still make sense in this case?

sethvargo: People can submit hours that are longer than a day (e.g. 72 hours), but if I think we want people to be able to do something more granular.

sethvargo (approved):

PR #42, opened 2023-06-15
feat: remove expired IAM bindings

sqin2019:


pkg/handler/iam_handler.go
```diff
@@ -219,3 +233,17 @@ func toBindingsMap(bs []*v1alpha1.Binding) map[string]map[string]struct{} {
 	}
 	return result
 }
+
+func expired(exp string) (bool, error) {
+	// An valid expiration expression will look like "request.time < timestamp('2023-06-14T19:55:03Z')".
+	matches := regexp.MustCompile(`request.time < timestamp\('([^']+)'\)`).FindStringSubmatch(exp)
```
sqin2019, line 239: see [here](https://github.com/abcxyz/access-on-demand/blob/main/pkg/handler/iam_handler.go#L194)

pkg/handler/iam_handler.go
```diff
@@ -133,7 +134,9 @@ func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolic
 		}
 
 		// Update the policy with new IAM binding additions.
-		updatePolicy(cp, p.Bindings, expiry)
+		if err := updatePolicy(cp, p.Bindings, expiry); err != nil {
+			return fmt.Errorf("failed to update IAM policy: %w", err)
+		}
```
sqin2019, line 139: The error will only occur when removing expired aod added IAM bindings. If the condition expression is not in a format of what we expected, then it will return error. In this PR, update the IAM policy of this resource will fail. the repo/gcp org/folder/project admin needs to check why that is happening, AOD will stop working for this resource until it is fixed.

pkg/handler/iam_handler.go
```diff
@@ -166,7 +169,17 @@ func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) {
 			result = append(result, cb)
 			continue
 		}
-		// TODO (#6): Remove expired bindings.
+
+		// Skip expired bindings.
+		expired, err := expired(cb.Condition.Expression)
```
drevell, line 174: Since AOD is an important emergency response tool, maybe we should allow the update to continue even if the expired policies can't be removed? It would be unfortunate if we had to block an emergency response that could have continued. Maybe we should find a way to alert about the problem but still continue?

pkg/handler/iam_handler.go
```diff
@@ -219,3 +233,17 @@ func toBindingsMap(bs []*v1alpha1.Binding) map[string]map[string]struct{} {
 	}
 	return result
 }
+
+func expired(exp string) (bool, error) {
+	// An valid expiration expression will look like "request.time < timestamp('2023-06-14T19:55:03Z')".
+	matches := regexp.MustCompile(`request.time < timestamp\('([^']+)'\)`).FindStringSubmatch(exp)
+	// Unable to find enough submatches.
+	if len(matches) < 2 {
+		return false, fmt.Errorf("expression %q does not match format %q", exp, "request.time < timestamp('2006-01-02T15:04:05Z07:00')")
```
capri-xiyue, line 242: Is it expected that we write hard-coded timestamp `2006-01-02T15:04:05Z07:00` here?

pkg/handler/iam_handler.go
drevell: It's a little safer to move the regex compilation to a global, like this:

```
var expirationRegex = regex.MustCompile(...)

func expired(...) {
```

This way, the regex will compiled when the program starts. This has two benefits:
 - it's automatically tested. If the regex fails to compile, then it will panic, and any tests that link to this package will fail.
 - it's only compiled once, not every time the function is called (not so important in this case where a few microseconds don't matter)

drevell (changes requested):

pkg/handler/iam_handler.go
sqin2019: Make sense, updating!

pkg/handler/iam_handler.go
sqin2019: Yeah, thanks for pointing it out, I thought about this as well. The reason I chose to stop here mainly because we don't expect this error, if there is, it probably means that the conditions are polluted and AOD is producing unexpected IAM updates, in this case someone/admin needs to take a look at this right away. I was wondering how to alert people about this problem and couldn't find a way except for using metrics (probably GMA or GCP metrics?). What is your suggestion?

pkg/handler/iam_handler.go
sqin2019: It is just a format example, will update and make it clearer.

pkg/handler/iam_handler.go
drevell: Maybe there could be alerting based on error logs? https://cloud.google.com/logging/docs/alerting/log-based-alerts .

Generally, I think "stop working until the problem is fixed" might not be the right attitude for an important operations tool.

sqin2019: Thank you, will look into it!

pkg/handler/iam_handler.go
drevell: To keep you unblocked, I'm fine if you want to just submit this PR

pkg/handler/iam_handler.go
drevell: I don't want to block this PR though, approved.

drevell (dismissed):

pkg/handler/iam_handler.go
sqin2019: Looks new commits will dismiss old approvals...

drevell (approved):

PR #43, opened 2023-06-16
fix: expire inputs

sqin2019:
To use [github.repository context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)(The owner and repository name. For example, octocat/Hello-World.) in the caller workflow, make it easier for caller. 
Below is an example caller
```
  close_pr:
    permissions:
      pull-requests: 'write'
    uses: 'abcxyz/access-on-demand/.github/workflows/expire.yml@main'
    with:
      repo: '${{ github.repository }}'
```

### DIFF
--- a/.github/workflows/expire.yml
+++ b/.github/workflows/expire.yml
@@ -25,8 +25,12 @@ jobs:
           result-encoding: 'string'
           retries: 3
           script: |-
-            const repoOwner = String('${{ inputs.repo }}').split('/')[0]
-            const repoName = String('${{ inputs.repo }}').split('/')[1]
+            const parts = String('${{ inputs.repo }}').split('/')
+            if (parts.length != 2) {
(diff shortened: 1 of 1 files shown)


.github/workflows/expire.yml
```diff
@@ -29,11 +25,13 @@ jobs:
           result-encoding: 'string'
           retries: 3
           script: |-
+            const repoOwner = String('${{ inputs.repo }}').split('/')[0]
```
drevell, line 28: Optional, if you feel like it: check that `${{ inputs.repo }}` contains a `/` and fail with a nice error messaage if not.

drevell (dismissed):

.github/workflows/expire.yml
sqin2019: Thanks, updated!

drevell (approved):

PR #48, opened 2023-06-21
feat: command-line tool request definition

sqin2019:


apis/v1alpha1/cli_request.go
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+// CLIRequest represents a request to run CLI commands.
+type CLIRequest struct {
+	// CLI name such as gcloud.
+	CLI string `yaml:"cli,omitempty"`
```
sqin2019, line 20: This is here because we may want to support other CLI like aws.

capri-xiyue:
Do you need to leverage this https://github.com/abcxyz/pkg/tree/main/cli when we implement the AOD cli?

sqin2019:
> Do you need to leverage this https://github.com/abcxyz/pkg/tree/main/cli when we implement the AOD cli?

Yes, for example,  `aod cli do -path "/path/to/<CLIRequestFileName>.yaml"` to handle CLIRequest.

capri-xiyue:
> > Do you need to leverage this https://github.com/abcxyz/pkg/tree/main/cli when we implement the AOD cli?
> 
> Yes, for example, `aod cli do -path "/path/to/<CLIRequestFileName>.yaml"` to handle CLIRequest.

If we use abcxyz/pkg/cli
do you still need to define such type CLIRequest struct? I thought the CLI request here should look like https://github.com/abcxyz/pmap/blob/main/pkg/cli/mapping_validate.go#L32 using abcxyz/pkg/cli

capri-xiyue (approved):

PR #50, opened 2023-06-22
feat: validate cli request

### LINKED ISSUE #47: gcloud request validation

sqin2019:
fix #47

### DIFF
--- a/apis/v1alpha1/validation.go
+++ b/apis/v1alpha1/validation.go
@@ -21,6 +21,11 @@ import (
 	"strings"
 )
 
+var (
+	defaultCLI              = "gcloud"
+	invalidCommandOperators = []string{"&", "|", ">", ">>"}
+)
+
 // ValidateIAMRequest checks if the IAMRequest is valid.
 func ValidateIAMRequest(r *IAMRequest) (retErr error) {
 	for _, s := range r.ResourcePolicies {
@@ -56,4 +61,40 @@ func ValidateIAMRequest(r *IAMRequest) (retErr error) {
 		}
 	}
 	return
+}
+
+// ValidateCLIRequest checks if the CLIRequest is valid.
+func ValidateCLIRequest(r *CLIRequest) (retErr error) {
+	// Set default CLI
+	if r.CLI == "" {
+		r.CLI = defaultCLI
+	}
+	// TODO (#49): support other CLIs.
+	if r.CLI != defaultCLI {
+		retErr = errors.Join(retErr, fmt.Errorf("CLI %q is not supported", r.CLI))
+	}
+
+	// Check if the do commands are valid.
+	for _, c := range r.Do {
+		if err := checkCommand(c); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("command %q is not valid: %w", c, err))
+		}
+	}
+
+	// Check if the cleanup commands are valid.
+	for _, c := range r.Cleanup {
+		if err := checkCommand(c); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("command %q is not valid: %w", c, err))
+		}
+	}
+	return retErr
+}
+
+func checkCommand(c string) error {
+	for _, o := range invalidCommandOperators {
+		if strings.Contains(c, o) {
+			return fmt.Errorf(`command contains invalid operators in %q`, invalidCommandOperators)
+		}
+	}
+	return nil
 }
--- a/apis/v1alpha1/validation.go
+++ b/apis/v1alpha1/validation.go
@@ -23,7 +23,11 @@ import (
 
 var (
 	defaultCLI              = "gcloud"
-	invalidCommandOperators = []string{"&", "|", ">", ">>"}
+	invalidCommandOperators = map[rune]struct{}{
+		'&': {},
+		'|': {},
+		'>': {},
+	}
 )
 
 // ValidateIAMRequest checks if the IAMRequest is valid.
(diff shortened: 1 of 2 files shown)


apis/v1alpha1/validation.go
```diff
@@ -57,3 +62,39 @@ func ValidateIAMRequest(r *IAMRequest) (retErr error) {
...
+}
+
+func checkCommand(c string) error {
+	for _, o := range invalidCommandOperators {
+		if strings.Contains(c, o) {
+			return fmt.Errorf(`command contains invalid operators in %q`, invalidCommandOperators)
+		}
+	}
+	return nil
+}
```
sethvargo, line 100: This is pretty inefficient (searches the string 4 times) and probably not complete. I would suggest:

```go
	invalidCommandOperators = map[string]struct{}{
    "&": {}, 
    "|": {}, 
    ">": {}, 
    ">>": {},
}
```

and then iterate over each character in the string and see if it's in the set of forbidden characters:

```go
for _, ch := range c {
  if _, ok := invalidCommandOperations[ch]; ok {
    return fmt.Errorf("invalid command operator %q", ch)
  }
}
```

sethvargo (changes requested):

apis/v1alpha1/validation.go
sqin2019: Got it, updating!

apis/v1alpha1/validation_test.go
```diff
@@ -200,3 +200,91 @@ func TestValidateIAMRequest(t *testing.T) {
...
+					"run jobs execute my-job2",
+				},
+				Cleanup: []string{
+					"run jobs executions delete my-execution1",
+					"run jobs executions delete my-execution2",
+				},
+			},
+		},
+		{
+			name: "success_without_cli",
```
capri-xiyue, line 227: nits: success_with_default_cli?

apis/v1alpha1/validation.go
```diff
@@ -57,3 +62,39 @@ func ValidateIAMRequest(r *IAMRequest) (retErr error) {
...
+	}
+	// TODO (#49): support other CLIs.
+	if r.CLI != defaultCLI {
+		retErr = errors.Join(retErr, fmt.Errorf("CLI %q is not supported", r.CLI))
+	}
+
+	// Check if the do commands are valid.
+	for _, c := range r.Do {
+		if err := checkCommand(c); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("command %q is not valid: %w", c, err))
```
capri-xiyue, line 84: nits, fmt.Errorf("do command %q is not valid: %w", c, err))

apis/v1alpha1/validation.go
```diff
@@ -57,3 +62,39 @@ func ValidateIAMRequest(r *IAMRequest) (retErr error) {
...
+		}
+	}
+
+	// Check if the cleanup commands are valid.
+	for _, c := range r.Cleanup {
+		if err := checkCommand(c); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("command %q is not valid: %w", c, err))
```
capri-xiyue, line 91: Same

apis/v1alpha1/validation.go
```diff
@@ -57,3 +66,39 @@ func ValidateIAMRequest(r *IAMRequest) (retErr error) {
...
+		}
+	}
+	return retErr
+}
+
+func checkCommand(c string) error {
+	for _, ch := range c {
+		if _, ok := invalidCommandOperators[ch]; ok {
+			return fmt.Errorf("invalid command operator %q", ch)
+		}
```
capri-xiyue, line 101: Currently, if multiple invalid command operators are matched, users have to change the command multiple time as each time the validation only returns one error related to one matched invalid operator

sqin2019: Right, can put the whole list in the error message. updating.

sqin2019: Since the list is short, we can stop early and provide all the invalid operators.

apis/v1alpha1/validation.go
```diff
@@ -57,3 +66,49 @@ func ValidateIAMRequest(r *IAMRequest) (retErr error) {
...
+func checkCommand(c string) error {
+	for _, ch := range c {
+		if _, ok := invalidCommandOperators[ch]; ok {
+			return fmt.Errorf("contains invalid command operators in %q", getMapKeys(invalidCommandOperators))
+		}
+	}
+	return nil
+}
+
+func getMapKeys(m map[rune]struct{}) []rune {
```
capri-xiyue, line 106: Why use `rune` here?

apis/v1alpha1/validation.go
```diff
@@ -21,6 +21,15 @@ import (
 	"strings"
 )
 
+var (
+	defaultCLI              = "gcloud"
+	invalidCommandOperators = map[rune]struct{}{
```
capri-xiyue, line 26: Why use rune instead of string here?

sqin2019: rune is char in go? wider range if using rune/char. for example I only need to check '&' for both & and &&. Also here we are iterating rune ```for _, ch := range c {```

capri-xiyue (approved):

apis/v1alpha1/validation.go
```diff
@@ -97,8 +97,18 @@ func ValidateCLIRequest(r *CLIRequest) (retErr error) {
...
+
+func getMapKeys(m map[rune]struct{}) []rune {
+	keys := make([]rune, len(m))
+	i := 0
+	for k := range invalidCommandOperators {
+		keys[i] = k
+		i++
+	}
+	return keys
+}
```
sethvargo, line 114: It's more idiomatic to do:

```suggestion
func getMapKeys(m map[rune]struct{}) []rune {
	keys := make([]rune, 0, len(m))
	for k := range invalidCommandOperators {
		keys = append(keys, k)
	}
	return keys
}
```

sethvargo: However, that being said, we probably don't need this. A better error message would be "disallowed character '&' at position 12:5", which we can get while iterating the string.

sqin2019: Xiyue and I discussed this and think that it might be better to show all the invalid operators in the command, so that people can resolve all the errors once and for all. I suggested this way so that we can stop early and provide more hint about what chars are not allowed. What do you think?

sethvargo:
```go
func checkCommand(c string) error {
	var merr error
	scanner := bufio.NewScanner(strings.NewReader(c))
	for row := 1; scanner.Scan(); row++ {
		line := scanner.Text()
		for col, r := range line {
			if _, ok := invalidCommandOperators[r]; ok {
				merr = errors.Join(merr, fmt.Errorf("disallowed command character %q at %d:%d", r, row, col))
			}
		}
	}
	return merr
}
```

That'll give output like:

```text
disallowed command character '&' at 3:18
disallowed command character '&' at 3:19
disallowed command character '|' at 3:29
disallowed command character '|' at 3:30
```

sqin2019: Yeah, I want to stop early though, the cli command is short, the invalidCommandOperators is also short, just a list telling user what is not allowed is enough, but again they are short so not much different performance wise.  Do we really need Scanner?

sethvargo: I don't think we can guarantee the command is going to be short. The scanner is for getting line number. Even if I have a really short command, it's going to be really annoying to find that character in a multi-line command.

sqin2019: Cool, updating!

sethvargo (approved):

PR #51, opened 2023-06-22
feat: CLI request handler

sqin2019:


pkg/handler/cli_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+// WithDefaultDebugMode sets the handler's stdout to os.Stdout.
+func WithDefaultDebugMode() CLIHandlerOption {
+	return func(h *CLIHandler) *CLIHandler {
+		h.stdout = os.Stdout
+		return h
+	}
+}
```
sethvargo, line 59: What is the difference between these?

sqin2019: Wanted to make os.Stdout a default option.

pkg/handler/cli_handler_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestCLIHandlerDo(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name           string
+		request        *v1alpha1.CLIRequest
+		expErrSubstr   string
```
capri-xiyue, line 35: handleErrSubStr?

pkg/handler/cli_handler_test.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+func TestCLIHandlerCleanup(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name           string
+		request        *v1alpha1.CLIRequest
+		expErrSubstr   string
```
capri-xiyue, line 105: Same

pkg/handler/cli_handler.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Cleanup runs the cleanup commands.
+func (h *CLIHandler) Cleanup(ctx context.Context, r *v1alpha1.CLIRequest) error {
+	return h.run(r.CLI, r.Cleanup)
+}
+
+func (h *CLIHandler) run(cli string, cmds []string) error {
+	for _, c := range cmds {
+		cmd := exec.Command(cli, c)
+		if h.stdout != nil {
```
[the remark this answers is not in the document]
sqin2019: yes, in general, we want to hide command outputs because some gcloud command outputs could be sensitive information. Only in debug mode when we want to check the command outputs we need to set stdout.

pkg/handler/cli_handler.go
```diff
@@ -0,0 +1,95 @@
...
+			cmd.Stdout = h.stdout
+		}
+		if h.stderr != nil {
```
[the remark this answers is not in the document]
sqin2019: ah...yeah, stderr will never be nil since we have default, can remove this check.

pkg/handler/cli_handler.go
```diff
@@ -0,0 +1,95 @@
...
+type CLIHandler struct {
+	// By default, stdout discards the command outputs, stderr is os.Stderr.
+	stdout, stderr io.Writer
+}
+
+// CLIHandlerOption is the option to set up an CLIHandler.
+type CLIHandlerOption func(h *CLIHandler) *CLIHandler
+
+// WithStderr sets the handler's stderr.
+func WithStderr(w io.Writer) CLIHandlerOption {
```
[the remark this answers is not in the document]
sqin2019: No, I set the stderr default in NewHandler so that we can always have stderr default.

capri-xiyue (approved):

pkg/handler/cli_handler.go
sqin2019: resolving for now, let me know if any concerns I will follow up later. Thanks!

PR #52, opened 2023-06-23
feat: aod cli do/cleanup command

sqin2019:


pkg/cli/iam_handle_test.go
```diff
@@ -128,7 +128,6 @@ policies:
 	cases := []struct {
 		name     string
 		args     []string
-		fileData []byte
```
sqin2019, line 131: Remove unused.

pkg/cli/cli_handle.go
```diff
@@ -0,0 +1,147 @@
...
+	if c.flagPath == "" {
+		return fmt.Errorf("path is required")
+	}
+
+	return c.do(ctx)
+}
+
+func (c *CLIHandleCommand) do(ctx context.Context) error {
+	// Read request from file path.
+	var req v1alpha1.CLIRequest
```
capri-xiyue, line 116: Use pointer here?

pkg/cli/root.go
```diff
@@ -42,6 +42,20 @@ var rootCmd = func() cli.Command {
...
+			"cli": func() cli.Command {
+				return &cli.RootCommand{
+					Name:        "cli",
+					Description: "Perform operations related to the CLI request",
+					Commands: map[string]cli.CommandFactory{
+						"do": func() cli.Command {
+							return &CLIHandleCommand{}
+						},
+						"cleanup": func() cli.Command {
+							return &CLIHandleCommand{Cleanup: true}
```
capri-xiyue, line 54: Should do and cleanup its its own command instead of using a bool to differentiate between do and cleanup? I understand do and cleanup share a lot of common logic, maybe we can refactor those logic as a util?

pkg/cli/cli_handle.go
sqin2019: ```var req *v1alpha1.CLIRequest``` will panic here ```panic: reflect: reflect.Value.Set using unaddressable value```
It needs to be either this ```var req v1alpha1.CLIRequest``` or ```req := &v1alpha1.CLIRequest{}```

sethvargo: I think the code as-is is correct. We're passing around the pointer (`&req`)

pkg/cli/root.go
sqin2019: I wanted to follow the pattern of IAMRequest. Handle command handles IAM/CLI request. the difference is CLI handle has two parts, do and cleanup. We can do a wrapper layer, ```"do": &CLIDoCommand{CLIHandleCommand{dofunc(ctx, *v1alpha1.CLIRequest) error}}```,  ```"cleanup": &CLICleanupCommand{CLIHandleCommand{cleanupfunc(ctx, *v1alpha1.CLIRequest) error}}```, but it is kinda redundant.

capri-xiyue: Is it possible that the CLI handle command will need to support more commands besides cleanup and do in the near future? If in the near future, we will only have do and cleanup, I'm fine with current structure.

sqin2019: Yes, based on the design cli requests only have do and cleanup, I also cannot think of any possible additional features besides these two. If needed, we can come back and refactor.

pkg/cli/cli_handle.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if c.flagPath == "" {
+		return fmt.Errorf("path is required")
+	}
+
+	return c.do(ctx)
+}
+
+func (c *CLIHandleCommand) do(ctx context.Context) error {
```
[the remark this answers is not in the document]
sqin2019: Done

pkg/cli/cli_handle.go
```diff
@@ -0,0 +1,147 @@
...
+// CLIHandleCommand handles CLI requests.
+type CLIHandleCommand struct {
+	cli.BaseCommand
+
+	flagPath string
+
+	flagDebug bool
+
+	// Run Cleanup instead of Do if true.
+	Cleanup bool
```
sethvargo, line 39: Sorry, I don't think this is the correct architecture. We should have a "do" command and a "cleanup" command. Branching logic based off of a boolean feels like a design smell.

sqin2019: In this case, what about make it a flag. so the commands are like do: `aod cli handle -path <path/to/file.yaml>`, cleanup: `aod cli handle -path <path/to/file.yaml> -cleanup`

capri-xiyue: We discussed similar question in https://github.com/abcxyz/access-on-demand/pull/52#discussion_r1240232927. Maybe it is still worthwhile to have a  `wrapper layer, "do": &CLIDoCommand{CLIHandleCommand{dofunc(ctx, *v1alpha1.CLIRequest) error}}, "cleanup": &CLICleanupCommand{CLIHandleCommand{cleanupfunc(ctx, *v1alpha1.CLIRequest) error}}` now?

sqin2019: I think this might be better in fact: https://github.com/abcxyz/access-on-demand/pull/55, thank you both! Let me know what you think.

PR #55, opened 2023-06-27
fix: cli usage info

sqin2019:

PR #56, opened 2023-06-27
feat: aod cli cleanup workflow

sqin2019:
hold until https://github.com/abcxyz/access-on-demand/pull/55 merges

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -86,4 +86,4 @@ jobs:
       - name: 'Handle cleanup'
         env:
           FILE_PATH: '${{ github.workspace }}/cli.yaml'
-        run: 'aod cli handle -path $FILE_PATH -cleanup'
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,98 @@
...
+        default: 'latest'
+        required: false
+      go_version:
+        description: 'The version of Golang.'
+        type: 'string'
+        default: '1.20'
+        required: false
+
+jobs:
+  approved:
```
sqin2019, line 37: An much easier way is to check ```if: '${{ github.event.pull_request.mergeable}}'``` instead of this approved job. But it won't work since all AOD requests will have a failure check to prevent them from merging.

sqin2019: I will try graphql instead, ref:https://github.com/orgs/community/discussions/24702

capri-xiyue (approved):

PR #61, opened 2023-07-10
fix: ignore errors if error occured when removing expired iam

### LINKED ISSUE #44: Investigate how improve IAM binding expire error alert
Follow up of https://github.com/abcxyz/access-on-demand/pull/42#discussion_r1231615773
Currently AOD will stop when condition expression format error occurs (in order to remove expired IAM bindings) during IAM policy update.  An improvement could be to continue IAM policy update and alert the error through other means.

sqin2019:
fix #44

### DIFF
--- a/pkg/handler/iam_handler.go
+++ b/pkg/handler/iam_handler.go
@@ -178,6 +178,8 @@ func updatePolicy(ctx context.Context, p *iampb.Policy, bs []*v1alpha1.Binding, 
 		// Skip expired bindings.
 		expired, err := expired(cb.Condition.Expression)
 		if err != nil {
+			// Continue policy update when there is error checking the AOD expiry.
+			// This type of error will be handled in a global cleanup job.
 			logger.Warnf("failed to check expiry: %w", err)
 		}
 		if expired {
--- a/pkg/handler/iam_handler.go
+++ b/pkg/handler/iam_handler.go
@@ -26,6 +26,7 @@ import (
 
 	"cloud.google.com/go/iam/apiv1/iampb"
 	"github.com/abcxyz/access-on-demand/apis/v1alpha1"
+	"github.com/abcxyz/pkg/logging"
 	"github.com/googleapis/gax-go/v2"
 	"github.com/sethvargo/go-retry"
 	"google.golang.org/genproto/googleapis/type/expr"
@@ -108,6 +109,7 @@ func (h *IAMHandler) Do(ctx context.Context, r *v1alpha1.IAMRequestWrapper) (nps
 }
 
 func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolicy, expiry time.Time) (*v1alpha1.IAMResponse, error) {
+	logger := logging.FromContext(ctx)
 	var iamC IAMClient
 	switch strings.Split(p.Resource, "/")[0] {
 	case "organizations":
@@ -139,11 +141,8 @@ func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolic
 			return fmt.Errorf("failed to get IAM policy: %w", err)
 		}
 
-		// TODO (#44): Continue to handle policy and alert updatePolicy error
-		// differently.
-		// Update the policy with new IAM binding additions.
 		if err := updatePolicy(cp, p.Bindings, expiry); err != nil {
-			return fmt.Errorf("failed to update IAM policy: %w", err)
+			logger.Warnf("encountered error while updating IAM policy for resource %q: %w", p.Resource, err)
 		}
 
 		// Set the new policy.
@@ -166,7 +165,7 @@ func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolic
 }
 
 // Remove expired bindings and add or update new bindings with expiration condition.
-func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) error {
+func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) (retErr error) {
 	// Convert new bindings to a role to unique bindings map.
 	bsMap := toBindingsMap(bs)
 	// Clean up current policy bindings.
@@ -181,8 +180,7 @@ func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) err
 		// Skip expired bindings.
 		expired, err := expired(cb.Condition.Expression)
 		if err != nil {
-			// Return error immediately since we don't expect this to fail.
-			return fmt.Errorf("failed to check expiry: %w", err)
+			retErr = errors.Join(retErr, fmt.Errorf("failed to check expiry: %w", err))
 		}
 		if expired {
 			continue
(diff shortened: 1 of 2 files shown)


pkg/handler/iam_handler.go
```diff
@@ -40,6 +40,9 @@ var (
 	expirationRegex = regexp.MustCompile(`request.time < timestamp\('([^']+)'\)`)
 )
 
+// updateIAMPolicy func that updates the given IAM policy.
+type updateIAMPolicy func(*iampb.Policy, []*v1alpha1.Binding, time.Time) error
```
sqin2019, line 44: Hi @yolocs , would appreciate your advice on below:
1. Using a func type to reduce duplicative code as much as possible, even though it passes unused parameters. 
2. I remember we want to make the logic of removing expired IAM (line 256) general, but reusability is not the goal in this PR.

sqin2019: note: global cleanup

pkg/handler/iam_handler.go
```diff
@@ -139,11 +141,8 @@ func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolic
 			return fmt.Errorf("failed to get IAM policy: %w", err)
 		}
 
-		// TODO (#44): Continue to handle policy and alert updatePolicy error
-		// differently.
-		// Update the policy with new IAM binding additions.
 		if err := updatePolicy(cp, p.Bindings, expiry); err != nil {
-			return fmt.Errorf("failed to update IAM policy: %w", err)
+			logger.Warnf("encountered error while updating IAM policy for resource %q: %w", p.Resource, err)
```
yolocs, line 145: It doesn't seem we differentiate errors from adding new AOD bindings vs. removing expired AOD bindings? We should only ignore "remove expired AOD binding error".

pkg/handler/iam_handler.go
```diff
@@ -181,8 +180,7 @@ func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) err
 		// Skip expired bindings.
 		expired, err := expired(cb.Condition.Expression)
 		if err != nil {
-			// Return error immediately since we don't expect this to fail.
-			return fmt.Errorf("failed to check expiry: %w", err)
+			retErr = errors.Join(retErr, fmt.Errorf("failed to check expiry: %w", err))
```
yolocs, line 183: This line particularly - we should ignore this error

yolocs (changes requested):

pkg/handler/iam_handler.go
sqin2019: updating.

pkg/handler/iam_handler.go
```diff
@@ -181,8 +178,7 @@ func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) err
 		// Skip expired bindings.
 		expired, err := expired(cb.Condition.Expression)
 		if err != nil {
-			// Return error immediately since we don't expect this to fail.
```
yolocs, line 184: No op - now I realized this is the only error it will return

yolocs (approved):

pkg/handler/iam_handler.go
sqin2019: Yeah, but it was confusing if it was logged there.

pkg/handler/iam_handler.go
```diff
@@ -139,12 +140,7 @@ func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolic
 			return fmt.Errorf("failed to get IAM policy: %w", err)
 		}
 
-		// TODO (#44): Continue to handle policy and alert updatePolicy error
-		// differently.
-		// Update the policy with new IAM binding additions.
-		if err := updatePolicy(cp, p.Bindings, expiry); err != nil {
-			return fmt.Errorf("failed to update IAM policy: %w", err)
-		}
+		updatePolicy(ctx, cp, p.Bindings, expiry)
```
sethvargo, line 143: I'm not sure we want to blindly ignore the error here - we should check that the error is specifically the one we want to ignore (either with grpcstatus or checking the error string).

pkg/handler/iam_handler.go
```diff
@@ -181,8 +178,9 @@ func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) err
 		// Skip expired bindings.
 		expired, err := expired(cb.Condition.Expression)
 		if err != nil {
-			// Return error immediately since we don't expect this to fail.
-			return fmt.Errorf("failed to check expiry: %w", err)
+			// Continue policy update when there is error checking the AOD expiry.
+			// This type of error will be handled in a global cleanup job.
+			logger.Warnf("failed to check expiry: %w", err)
```
sethvargo, line 183:
```suggestion
			logger.Warnw("failed to check expiry", "error", err)
```

pkg/handler/iam_handler.go
yolocs: In this case we do want to ignore blindly. We don't want to fail just because of cleanup failure. Maybe a comment will help

sethvargo: Wouldn't a cleanup failure mean we have overzealous permissions somewhere?

yolocs: This is best effort to clean up old AOD IAM bindings **when applying a new AOD IAM binding**.

As a catch-it-all mechanism, at the moment we plan to do a global scan of AOD bindings near [guardian drift detection](https://github.com/abcxyz/guardian/issues/145) and clean up any findings.

We still plan to do the clean up on AOD PR close. But because we allow multiple iterations in an AOD PR, it's possible one asks for entirely different permission in different iterations. E.g. I can start with a commit to ask for BQ permission then switch over to pubsub permission. When we do the clean up, if we only look at the latest commit, we will only clean up the pubsub permission but not the BQ one. Ideally we should look at all **approved commits** in this PR to clean up. That's much challenging to do.

sqin2019: Here is the ongoing PR that is relevant: https://github.com/abcxyz/guardian/pull/148/files

pkg/handler/iam_handler.go
```diff
@@ -181,8 +182,9 @@ func updatePolicy(p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) err
 		// Skip expired bindings.
 		expired, err := expired(cb.Condition.Expression)
 		if err != nil {
-			// Return error immediately since we don't expect this to fail.
-			return fmt.Errorf("failed to check expiry: %w", err)
+			// Continue policy update when there is error checking the AOD expiry.
+			// This type of error will be handled in a global cleanup job.
```
yolocs, line 186:
```suggestion
			// Continue policy update when there is error checking the AOD expiry.
			// Cleaning up expired AOD bindings here is best effort.
			// We rely on a separate process to clean up AOD bindings.
```

yolocs (commented):
nit

yolocs (approved):

pkg/handler/iam_handler.go
sqin2019: Close it for now, please reopen if any concerns.

PR #63, opened 2023-07-11
fix: only handles IAM PR with required approval

sqin2019:
Fixes 57, to use this change, aod-template needs to add job permission `pull-requests: 'read'`

### DIFF
--- a/.github/workflows/handle.yml
+++ b/.github/workflows/handle.yml
@@ -47,18 +47,14 @@ jobs:
     outputs:
       REVIEW_DECISION: '${{ steps.get_review_decision.outputs.REVIEW_DECISION }}'
     steps:
-      - id: 'repo_name'
-        env:
-          REPO: '${{ github.repository }}'
-        run: 'echo "REPO_NAME=${REPO##*/}" >> $GITHUB_OUTPUT'
       - id: 'get_review_decision'
         env:
-          OWNER: '${{ github.repository_owner }}'
-          REPO_NAME: '${{ steps.repo_name.outputs.REPO_NAME }}'
-          PR_NUMBER: '${{ github.event.pull_request.number }}'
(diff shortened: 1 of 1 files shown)


sqin2019:
Hold until we explore the alternative flow.

.github/workflows/handle.yml
```diff
@@ -43,9 +42,38 @@ env:
...
 jobs:
+  review_status:
+    runs-on: 'ubuntu-latest'
+    outputs:
+      REVIEW_DECISION: '${{ steps.get_review_decision.outputs.REVIEW_DECISION }}'
+    steps:
+      - id: 'repo_name'
+        env:
+          REPO: '${{ github.repository }}'
+        run: 'echo "REPO_NAME=${REPO##*/}" >> $GITHUB_OUTPUT'
```
yolocs, line 53: Can we inline this in this next step?

.github/workflows/handle.yml
```diff
@@ -43,9 +42,38 @@ env:
...
+      - id: 'get_review_decision'
+        env:
+          OWNER: '${{ github.repository_owner }}'
+          REPO_NAME: '${{ steps.repo_name.outputs.REPO_NAME }}'
+          PR_NUMBER: '${{ github.event.pull_request.number }}'
```
yolocs, line 58: Consider inline the owner and PR number

.github/workflows/handle.yml
```diff
@@ -43,9 +42,38 @@ env:
...
+          GH_TOKEN: '${{ github.token }}'
```
yolocs, line 59: If this is needed, add a comment explain why and hopefully reference.

yolocs (commented):
nits

yolocs (approved):

PR #67, opened 2023-07-13
fix: referencing github env vars and typo

sqin2019:
not sure why, but it seems causing args parsing issues see https://github.com/abcxyz/aod-template/actions/runs/5547751120/jobs/10129816341?pr=25

----> found the cause, it not due to ${}, but due to typo...

### DIFF
--- a/.github/workflows/handle.yml
+++ b/.github/workflows/handle.yml
@@ -127,8 +127,8 @@ jobs:
           IAM_FILE_PATH: '${{ github.workspace }}/iam.yaml'
           START_TIME: '${{ github.event.review.submitted_at }}'
         run: |
-          touch $IAM_ERROR_FILENAME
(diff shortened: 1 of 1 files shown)


.github/workflows/handle.yml
```diff
@@ -127,16 +127,16 @@ jobs:
           IAM_FILE_PATH: '${{ github.workspace }}/iam.yaml'
           START_TIME: '${{ github.event.review.submitted_at }}'
         run: |
-          touch ${IAM_ERROR_FILENAME}
-          aod iam handle -path ${FILE_PATH} -duration ${DURATION} -start-time ${START_TIME} 2> ${IAM_ERROR_FILENAME}
+          touch $IAM_ERROR_FILENAME
+          aod iam handle -path $IAM_FILE_PATH -duration $DURATION -start-time $START_TIME 2> $IAM_ERROR_FILENAME
```
sethvargo, line 131: I'd recommend using `$${}` instead

sethvargo (approved):

PR #69, opened 2023-07-14
fix: note workflow - remove incorrect and unecessary condition

sqin2019:
for this condition to work need to checkout the triggering branch. We can add this condition in the triggering workflow instead
```
on:
  pull_request:
    types:
      - 'opened'
    paths:
      - 'iam.yaml'
```

### DIFF
--- a/.github/workflows/note.yml
+++ b/.github/workflows/note.yml
@@ -33,7 +33,6 @@ jobs:
     name: 'Add AOD Note'
     steps:
       - name: 'Add AOD Note'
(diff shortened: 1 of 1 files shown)


yolocs (approved):
Which part is incorrect? It probably won't hurt to have it?

sqin2019:
> Which part is incorrect? It probably won't hurt to have it?

it won't be able to search the files in the pull request unless it checkouts the triggering branch, which is a annoying/time-consuming step, so I figured why not remove it and it could be easily addressed in the calling workflow.

PR #75, opened 2023-07-19
refactor: tool do and clean commands

sqin2019:


pkg/cli/tool_base.go
```diff
@@ -0,0 +1,100 @@
...
+// ToolBaseCommand is the base command for handling tool requests.
+type ToolBaseCommand struct {
+	cli.BaseCommand
+
+	flagPath string
+
+	flagDebug bool
+
+	// testTool is used for testing only.
+	testTool string
```
yolocs, line 37: Using `echo` as the test tool doesn't seem right. See my suggestion in the other comment

pkg/cli/tool_base.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if err := v1alpha1.ValidateToolRequest(&req); err != nil {
+		return nil, nil, fmt.Errorf("failed to validate %T: %w", &req, err)
+	}
+
+	opts := []handler.ToolHandlerOption{handler.WithStderr(c.Stderr())}
+	if c.flagDebug {
+		opts = append(opts, handler.WithDebugMode(c.Stdout()))
+	}
+	h := handler.NewToolHandler(ctx, opts...)
```
yolocs, line 92: Make an interface of the handler so it can be faked.

yolocs (changes requested):

pkg/cli/tool_base.go
sqin2019: It will make testing the opts harder though, debug and error outputs.

yolocs (approved):

PR #84, opened 2023-07-27
feat: output actions in cli

### LINKED ISSUE #79: Have better output
`-verbose` flag will have the commands to output what actions it has taken. E.g.

```
Applied the following IAM bindings:

[The request]
```

```
Executed the following commands:

gcloud xxx
```

Having this standard output will make posting comments easier - it no longer needs to read the request file.

And maybe as a separate issue to have a `-debug` flag to print out more details. For IAM, it will be the whole IAM policies set. For gcloud, it will print out the gcloud output.

sqin2019:
fix #79

### DIFF
--- a/pkg/cli/iam_handle.go
+++ b/pkg/cli/iam_handle.go
@@ -24,6 +24,7 @@ import (
 	"github.com/abcxyz/access-on-demand/pkg/requestutil"
 	"github.com/abcxyz/pkg/cli"
 	"github.com/posener/complete/v2/predict"
+	"gopkg.in/yaml.v3"
 
 	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
 )
@@ -45,6 +46,8 @@ type IAMHandleCommand struct {
 
 	flagStartTime time.Time
 
+	flagVerbose bool
+
 	// testHandler is used for testing only.
 	testHandler iamHandler
 }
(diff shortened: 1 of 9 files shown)


pkg/cli/iam_handle.go
```diff
@@ -178,7 +192,20 @@ func (c *IAMHandleCommand) handleIAM(ctx context.Context) error {
 	if _, err := h.Do(ctx, reqWrapper); err != nil {
 		return fmt.Errorf("failed to handle IAM request: %w", err)
 	}
-	c.Outf("Successfully handled IAM request")
+
+	if c.flagVerbose {
+		enc := yaml.NewEncoder(c.Stdout())
```
sethvargo, line 197: Since this is repeated in multiple commands, and since it could use some dedicated unit tests, I'd recommend making a helper like `encodeYAML(w io.Writer)` that encapsulates this logic for the cli. It looks like you already kind of did that with `output`, but it's a slightly different API.

pkg/cli/iam_handle.go
```diff
@@ -178,7 +192,20 @@ func (c *IAMHandleCommand) handleIAM(ctx context.Context) error {
...
+		enc.SetIndent(2)
+		if err := enc.Encode(reqWrapper); err != nil {
+			return fmt.Errorf("failed to encode to yaml: %w", err)
+		}
+
+		if err := enc.Close(); err != nil {
+			return fmt.Errorf("failed to close yaml encoder: %w", err)
+		}
+	} else {
+		c.Outf("Successfully handled IAM request")
```
yolocs, line 207: Why skipping output this if verbose?

pkg/cli/iam_handle.go
sqin2019: Sure, updating.

pkg/cli/iam_handle.go
sqin2019: Included now.

pkg/cli/iam_handle.go
sqin2019: Add helper in util.go, I wonder if we need dedicated unit tests for private helpers?

yolocs:
Shall we consider https://github.com/abcxyz/access-on-demand/issues/79#issuecomment-1654614906?

sethvargo (approved):

yolocs (approved):

PR #89, opened 2023-08-10
fix: tool command cannot handle multi arg commands

sqin2019:


pkg/handler/tool_handler.go
```diff
@@ -93,9 +96,26 @@ func (h *ToolHandler) run(tool string, cmds []string) error {
...
 		// Empty line in between commands.
-		if i < (len(cmds) - 1) {
+		if h.stdout != nil && i < (len(cmds)-1) {
 			fmt.Fprintf(cmd.Stdout, "\n")
 		}
 	}
 	return nil
 }
+
+func splitArgs(str string) []string {
```
yolocs, line 106: As we discussed offline, we can try run `bash -c "gcloud xxx"` instead. Like you said, that's likely safer

pkg/handler/tool_handler.go
```diff
@@ -84,38 +82,21 @@ func (h *ToolHandler) Cleanup(ctx context.Context, r *v1alpha1.ToolRequest) erro
 
 func (h *ToolHandler) run(tool string, cmds []string) error {
 	for i, c := range cmds {
-		args := splitArgs(c)
-		cmd := exec.Command(tool, args...)
+		script := fmt.Sprintf("%s %s", tool, c)
+		cmd := exec.Command("bash", "-c", script)
```
yolocs, line 86: Comment here and in the CLI desc that tool only works where bash is available.

pkg/handler/tool_handler.go
```diff
@@ -84,38 +82,21 @@ func (h *ToolHandler) Cleanup(ctx context.Context, r *v1alpha1.ToolRequest) erro
 
 func (h *ToolHandler) run(tool string, cmds []string) error {
 	for i, c := range cmds {
-		args := splitArgs(c)
-		cmd := exec.Command(tool, args...)
+		script := fmt.Sprintf("%s %s", tool, c)
```
yolocs, line 85:
```suggestion
		toolCmd := fmt.Sprintf("%s %s", tool, c)
```

pkg/handler/tool_handler.go
yolocs: Also comment here about why we use bash.

yolocs (commented):
nit

pkg/handler/tool_handler.go
sqin2019: ack

yolocs (approved):

PR #93, opened 2023-08-15
feat: add custom condition description for iam handle

sqin2019:
In order to do AOD integration test, I compared two approaches, and decided to use 1.

1.  using a unique condition title for integration, the pitfall is that we will need an extra job to cleanup expired bindings as our best effort cleanup only cleanup bindings with the same condition title, I am thinking to use our tool command to do clean up after each integration. 
    - cleanup commd: gcloud projects remove-iam-policy-binding [PROJECT_ID](https://cloud.google.com/sdk/gcloud/reference/projects/remove-iam-policy-binding#PROJECT_ID) [--member](https://cloud.google.com/sdk/gcloud/reference/projects/remove-iam-policy-binding#--member)=PRINCIPAL [--role](https://cloud.google.com/sdk/gcloud/reference/projects/remove-iam-policy-binding#--role)=ROLE [[--all](https://cloud.google.com/sdk/gcloud/reference/projects/remove-iam-policy-binding#--all)     | [--condition](https://cloud.google.com/sdk/gcloud/reference/projects/remove-iam-policy-binding#--condition)=[KEY=VALUE,…] 
2. using unique condition description but the same condition title, we don't need to do extra cleanup, but it won't be able to run integration tests parallelly as AOD will overwrite dup bindings...

### DIFF
--- a/pkg/cli/iam_handle.go
+++ b/pkg/cli/iam_handle.go
@@ -47,8 +47,9 @@ type IAMHandleCommand struct {
 
 	flagVerbose bool
 
-	// flagConditionDescription is required for integration test.
-	flagConditionDescription string
+	// Optional custom condition title for IAM bindings expiration, required for
+	// integration test.
+	flagCustomConditionTitle string
 
 	// testHandler is used for testing only.
 	testHandler iamHandler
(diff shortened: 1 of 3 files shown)


pkg/cli/iam_handle.go
```diff
@@ -106,6 +109,14 @@ func (c *IAMHandleCommand) Flags() *cli.FlagSet {
 		Usage:   `Turn on verbose mode to print updated IAM policies. Note that it may contain sensitive information`,
 	})
 
+	f.StringVar(&cli.StringVar{
+		Name:    "condition-description",
+		Target:  &c.flagConditionDescription,
+		Hidden:  true,
+		Example: "example-description",
```
sethvargo, line 116: Is there a better example?

sqin2019: Sure, will update!

yolocs:
I like this approach!

yolocs (approved):

Issue #104, opened 2023-08-23
IAM cleanup


sqin2019:
Cleanup IAM when request PR is closed.

PR #108, opened 2023-08-24
refactor: iam expire func

sqin2019:
part of #104

### DIFF
--- a/pkg/handler/iam_handler.go
+++ b/pkg/handler/iam_handler.go
@@ -61,6 +61,9 @@ type IAMClient interface {
 	SetIamPolicy(context.Context, *iampb.SetIamPolicyRequest, ...gax.CallOption) (*iampb.Policy, error)
 }
 
+// updatePolicy updates the given IAM policy.
+type updatePolicy func(context.Context, *iampb.Policy, []*v1alpha1.Binding, time.Time)
+
 // Option is the option to set up an IAMHandler.
 type Option func(h *IAMHandler) (*IAMHandler, error)
 
@@ -106,11 +109,29 @@ func NewIAMHandler(ctx context.Context, organizationsClient, foldersClient, proj
 	return h, nil
 }
 
+// Cleanup removes expired IAM bindings added by AOD from the IAM policies of the resources in the request.
+func (h *IAMHandler) Cleanup(ctx context.Context, r *v1alpha1.IAMRequestWrapper) (nps []*v1alpha1.IAMResponse, retErr error) {
+	expiry := r.StartTime.Add(r.Duration)
+	for _, p := range r.ResourcePolicies {
+		np, err := h.handlePolicy(ctx, p, expiry, h.cleanupExpiredBindings)
+		if err != nil {
+			retErr = errors.Join(
+				retErr,
+				fmt.Errorf("failed to handle policy cleanup for resource %s: %w", p.Resource, err),
+			)
+		}
+		if np != nil {
+			nps = append(nps, np)
+		}
+	}
+	return
+}
+
 // Do removes expired or duplicative IAM bindings added by AOD and adds requested IAM bindings to current IAM policy.
 func (h *IAMHandler) Do(ctx context.Context, r *v1alpha1.IAMRequestWrapper) (nps []*v1alpha1.IAMResponse, retErr error) {
 	expiry := r.StartTime.Add(r.Duration)
 	for _, p := range r.ResourcePolicies {
-		np, err := h.handlePolicy(ctx, p, expiry)
+		np, err := h.handlePolicy(ctx, p, expiry, h.addAndCleanupPolicy)
 		if err != nil {
 			retErr = errors.Join(
 				retErr,
@@ -124,7 +145,7 @@ func (h *IAMHandler) Do(ctx context.Context, r *v1alpha1.IAMRequestWrapper) (nps
 	return
 }
 
-func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolicy, expiry time.Time) (*v1alpha1.IAMResponse, error) {
+func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolicy, expiry time.Time, updateFunc updatePolicy) (*v1alpha1.IAMResponse, error) {
 	var iamC IAMClient
 	switch strings.Split(p.Resource, "/")[0] {
 	case "organizations":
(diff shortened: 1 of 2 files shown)


pkg/handler/iam_handler.go
```diff
@@ -106,11 +109,29 @@ func NewIAMHandler(ctx context.Context, organizationsClient, foldersClient, proj
 	return h, nil
 }
 
+// Cleanup removes expired IAM bindings added by AOD from the IAM policies of the resources in the request.
+func (h *IAMHandler) Cleanup(ctx context.Context, r *v1alpha1.IAMRequest) (nps []*v1alpha1.IAMResponse, retErr error) {
```
sqin2019, line 113: Though we have talked about this, but I wonder what will be the best approach here @yolocs :
1. when cleanup triggered on a request, cleanups all expired the AOD binding in the policy. 
2. only cleanups bindings on the request that are expired.
3. only cleanups bindings on the request no matter if it is expired.

yolocs: I think it should be a combination of 1 and 3. It's safe to clean up all expired bindings. For 3, worst case scenario, a user needs to reopen the PR or open a new one.

pkg/handler/iam_handler.go
```diff
@@ -106,11 +109,29 @@ func NewIAMHandler(ctx context.Context, organizationsClient, foldersClient, proj
...
+	for _, p := range r.ResourcePolicies {
+		// time.Now() is a dummy parameter used to match the function signature.
+		np, err := h.handlePolicy(ctx, p, time.Now(), h.cleanupExpiredBindings)
```
yolocs, line 116: I don't think we should ignore cleanup error if it's an explicit cleanup call.

pkg/handler/iam_handler.go
```diff
@@ -181,37 +202,33 @@ func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolic
...
 
-// Remove expired bindings and add or update new bindings with expiration condition.
-func (h *IAMHandler) updatePolicy(ctx context.Context, p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) {
-	logger := logging.FromContext(ctx)
+// addAndCleanupPolicy adds new bindings with expiration condition, it also does
+// best effort cleanup which removes any expired AOD bindings, however any
+// errors encounterred during removal will be ignored and policy update for the
+// request will continue. Removal errors should be handled separately such as
+// in a global IAM cleanup.
+func (h *IAMHandler) addAndCleanupPolicy(ctx context.Context, p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) {
```
yolocs, line 210: For simplicity, I'd just call this `addPolicyBindings`

pkg/handler/iam_handler.go
```diff
@@ -181,37 +202,33 @@ func (h *IAMHandler) handlePolicy(ctx context.Context, p *v1alpha1.ResourcePolic
...
+	// Cleanup policy.
+	h.cleanupExpiredBindings(ctx, p, bs, expiry)
```
yolocs, line 212: This func should return an error and log warn here instead.

pkg/handler/iam_handler.go
```diff
@@ -247,6 +264,38 @@ func (h *IAMHandler) updatePolicy(ctx context.Context, p *iampb.Policy, bs []*v1
 	p.Version = 3
 }
 
+// cleanupExpiredBindings does best effort cleanup which removes any expired
+// AOD bindings, however any errors encounterred during removal will be ignored
+// Removal errors should be handled separately such as in a global IAM cleanup.
+// bs and expiry parameters are not used, they are here to match the
+// updatePolicy func signature.
+func (h *IAMHandler) cleanupExpiredBindings(ctx context.Context, p *iampb.Policy, bs []*v1alpha1.Binding, expiry time.Time) {
```
yolocs, line 272:
```suggestion
func (h *IAMHandler) cleanupExpiredBindings(ctx context.Context, p *iampb.Policy, _ []*v1alpha1.Binding, _ time.Time) {
```

yolocs (changes requested):

pkg/handler/iam_handler.go
sqin2019: Updated.

pkg/handler/iam_handler.go
sqin2019: Updated

pkg/handler/iam_handler.go
sqin2019: Updated.

pkg/handler/iam_handler.go
```diff
@@ -245,6 +239,59 @@ func (h *IAMHandler) updatePolicy(ctx context.Context, p *iampb.Policy, bs []*v1
...
+		}
+
+		// Check the expiration using expression directly instead of the given
+		// expiry as they may be different.
+		expired, err := expired(b.Condition.Expression)
+		if err != nil {
+			// Continue policy update when there is error checking the AOD expiry and
+			// return the errors.
+			// Cleaning up expired AOD bindings here is best effort.
+			// We rely on a separate process to clean up AOD bindings.
```
yolocs, line 266: Remove?

pkg/handler/iam_handler_test.go
```diff
@@ -1102,6 +1102,507 @@ func TestDo(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			wantFoldersPolicy:  &iampb.Policy{},
+			wantProjectsPolicy: &iampb.Policy{},
+			wantErrSubstr:      "failed to check expiry",
+		},
+		{
+			name: "ignore_non-AOD_bindings",
```
yolocs, line 1382: How about adding a case for ignoring non-requester bindings?

pkg/handler/iam_handler_test.go
```diff
@@ -1102,6 +1102,507 @@ func TestDo(t *testing.T) {
...
+				},
+			},
+			wantErrSubstr:           "Set IAM policy encountered error: Internal Server Error",
+			wantOrganizationsPolicy: &iampb.Policy{},
+			wantFoldersPolicy: &iampb.Policy{
+				Bindings: []*iampb.Binding{},
+			},
+			wantProjectsPolicy: &iampb.Policy{},
+		},
+	}
```
yolocs, line 1548: How about adding a case for mixing multiple bindings together? Expired, not expired, requester bindings, non-requester bindings, etc.

sqin2019: Added.

pkg/handler/iam_handler_test.go
sqin2019: Added

yolocs (approved):

Issue #115, opened 2023-08-31
Issues in e2e


sqin2019:
1. Deleting the branch right after PR is closed will lead to cleanup failure, since in the workflow it will fail to check out the triggering branch. 

Fix: Requester needs to delete the branch after the workflow checks out the trigger branch. 

2. Automatically closed PR does not trigger cleanup workflow. (it does if PR is closed manually)

Note: If it is fixed, we need to wait after PR is closed and before branch is deleted so that cleanup workflow can be run successfully.

sqin2019:
1. fix: update the workflow to close the branch at the end. 
2. cause: https://github.com/orgs/community/discussions/25565#discussioncomment-3248330
fix: update the workflow to call the cleanup workflow before deleting the branch.

sqin2019:
Remaining issue from 2, if expire job closes a PR that is not approved yet, the cleanup won't be triggered therefore branch is not deleted. Need to ensure branch is deleted after PR is closed.

PR #120, opened 2023-09-08
fix: cleanup should be triggered when PR is automatically closed

sqin2019:
part of #115

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -37,11 +37,11 @@ on:
         default: '1.21'
         required: false
       pr_number:
-        description: "The number of the pull request that contains the AOD requests."
+        description: 'The number of the pull request that contains the AOD requests.'
         type: 'number'
         required: true
       branch:
-        description: "The branch of the pull request."
+        description: 'The branch of the pull request.'
         type: 'string'
         required: true
 
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -36,6 +36,14 @@ on:
         type: 'string'
         default: '1.21'
         required: false
+      pr_number:
+        description: "The number of the pull request that contains the AOD requests."
+        type: 'number'
+        required: true
+      branch:
+        description: "The branch of the pull request."
+        type: 'string'
+        required: true
 
 env:
   IAM_ERROR_FILENAME: 'iam_error.txt'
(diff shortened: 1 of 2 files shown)


.github/workflows/cleanup.yml
```diff
@@ -36,6 +36,14 @@ on:
         type: 'string'
         default: '1.21'
         required: false
+      pr_number:
```
sqin2019, line 39: Need to make these two inputs as the end expire workflow needs to trigger this workflow, cannot make ${{ github.event.pull_request.number }} as default as it will fail if trigger is not pull_request event even with pr_number provided as input.

.github/workflows/expire.yml
```diff
@@ -27,22 +27,27 @@ on:
         type: 'number'
         default: 24
         required: false
+    outputs:
```
sqin2019, line 30: This will be used in end caller workflow to run cleanup parallelly.

.github/workflows/cleanup.yml
```diff
@@ -280,12 +288,12 @@ jobs:
             await github.rest.issues.createComment({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              issue_number: ${{ github.event.pull_request.number }},
+              issue_number: ${{ inputs.pr_number }},
               body: `**\`Access on Demand\`** - 🟦 **\`IAM\`** request not found, skip cleanup.`,
             });
       - name: 'Delete Branch'
         # Only delete branch when cleanups are skipped or successfully handled.
-        if: '${{ always() && steps.cleanup_iam.outcome != ''failure'' && steps.cleanup_tool.outcome != ''failure'' }}'
```
sqin2019, line 288: Removed always() as we should't delete the branch if any of the previous steps failed.

.github/workflows/expire.yml
```diff
@@ -110,15 +117,10 @@ jobs:
...
-                  ref: `heads/${ref}`,
-                });
+                expiredPulls.push({pullNumber: pull['number'], pullBranch: pull['head']['ref']})
               } else {
                 core.info(`Skipping #${pull['number']} (${pull['title']}) - not an AOD request`);
               }
-            });
+            }));
+
+            return expiredPulls;
```
yolocs, line 126: Is the plan to for each expired pulls to call cleanup workflow?

yolocs (approved):

.github/workflows/expire.yml
sqin2019: yes, we can use github matrix to run them parallelly, per Github doc, a matrix will generate a maximum of 256 jobs per workflow run.

PR #135, opened 2023-10-02
chore: add create tag workflow

### LINKED ISSUE #128: Enforce required reviews on releases
Investigate [GitHub deployment](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments), use it for the release workflow so that we can enforce required approvals.

sqin2019:
Won't work until pkg and minty change merged

Parallel TODOs
- need to create the `tag` deployment env.
- need to set repo vars for token minter inputs. -> done as they are set at the org level

fix #128 

The workflow page will look similar to (the outputs section result below is fixed):
image

### DIFF
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -1,0 +1,77 @@
+# Copyright 2023 The Authors (see AUTHORS file)
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#      http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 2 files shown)


.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,77 @@
...
+        required: false
+      message:
+        description: 'Message for the tag. Default will be the tag name.'
+        type: 'string'
+        required: false
+      annotated_tag:
+        description: 'This is an annotated tag.'
+        type: 'boolean'
+        default: true
+        required: false
```
yolocs, line 36: Let's remove this input if it's not needed.

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,77 @@
...
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The name of the tag to be created.'
+        type: 'string'
+        required: true
+      branch:
+        description: 'The branch of the head commit to create tag on. Default will be the repo default branch.'
+        type: 'string'
+        required: false
```
yolocs, line 27: Let's remove the input if it's not used.

yolocs (commented):
It's reasonable for the underlying reusable workflow to provide more flexibility. But here if we know we always create release on main with annotated tag, we shouldn't have those be inputs.

yolocs (approved):

PR #154, opened 2024-04-16
feat: add breakglass capabilities

verbanicm:
This adds the ability to create AoD requests for breakglass scenarios. By adding the `aod-breakglass` label to an AoD pr, it will automatically be applied.

This approach was selected over requiring an issue link as this creates less toil. The justification can be provided in the PR body rather than in a separate issue which just adds an extra step on a separate screen.

### DIFF
--- a/.github/workflows/handle.yml
+++ b/.github/workflows/handle.yml
@@ -50,9 +50,17 @@ jobs:
     permissions:
       pull-requests: 'read'
     outputs:
-      REVIEW_DECISION: '${{ steps.get_review_decision.outputs.REVIEW_DECISION }}'
+      REVIEW_DECISION: '${{ steps.get_breakglass_decision.outputs.REVIEW_DECISION || steps.get_review_decision.outputs.REVIEW_DECISION }}'
     steps:
+      - id: 'get_breakglass_decision'
+        if: |-
+          contains(github.event.pull_request.labels.*.name, 'aod-breakglass')
+        run: |
+          echo REVIEW_DECISION=BREAKGLASS >> $GITHUB_OUTPUT
+
       - id: 'get_review_decision'
+        if: |-
+          !contains(github.event.pull_request.labels.*.name, 'aod-breakglass')
         env:
           # Set the GH_TOKEN environment variable to use GitHub CLI in a GitHub Actions workflow.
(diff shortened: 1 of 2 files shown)


sailorlqh (approved):

yolocs (approved):

.github/workflows/handle.yml
```diff
@@ -50,9 +50,17 @@ jobs:
     permissions:
       pull-requests: 'read'
     outputs:
-      REVIEW_DECISION: '${{ steps.get_review_decision.outputs.REVIEW_DECISION }}'
+      REVIEW_DECISION: '${{ steps.get_breakglass_decision.outputs.REVIEW_DECISION || steps.get_review_decision.outputs.REVIEW_DECISION }}'
     steps:
+      - id: 'get_breakglass_decision'
+        if: |-
+          contains(github.event.pull_request.labels.*.name, 'aod-breakglass')
```
sqin2019, line 57: Nit: should it be github.event.label.name == 'aod-breakglass' ? Since handle will now also be triggered by labeling event.

Thank you for making this change!

.github/workflows/note.yml
```diff
@@ -46,6 +54,8 @@ jobs:
     name: 'Add AOD Note'
     steps:
       - name: 'Add AOD Note'
+        if: |-
+          !contains(github.event.pull_request.labels.*.name, 'aod-breakglass')
```
sqin2019, line 58: Same here, otherwise random labels can trigger this workflow too. Also, just curious can we select path for label events similar to [here](https://github.com/abcxyz/aod-template/blob/main/.github/workflows/note.yml#L22)?

sqin2019 (commented):
Thank you! Are we going to validate the justification?

.github/workflows/handle.yml
verbanicm: I think this syntax will stil work and it will work across event types. There might be an issue here though because pull request and pull request review show up as different status checks....let me dig in after I merge this and test it on one of our repos.

.github/workflows/note.yml
verbanicm: I think for consistency, we should just use this as its the same for both pull_request and pull_request_review events.

.github/workflows/handle.yml
sqin2019: Sounds like a plan!

PR #187, opened 2026-02-25
fix: handle workflow to checkout the approved commit

sqin2019:
use SHA so that it checks out the snapshot that triggered the workflow


crwilcox (approved):
👍

bradegler (approved):

sqin2019:
tested, it is now checking out the SHA:
Screenshot 2026-02-26 at 10 09 41 AM

